PR #12, opened 2026-07-03
feat(examples): GPU advection benchmark — stall/ceiling metrics + cloud synthetic store

emfdavid:
## GPU advection benchmark — round 1 (backlog #9)

Instruments the advection torch training loop with a self-referential loader-health
metric set and runs it on GPU against WeatherBench2 ERA5 and a large synthetic store
on GCS. Grounds the M2 `device_transform` design in measured behavior before building it.

### What's added
- **`examples/_forecast_metrics.py`** — framework-neutral collector: `ForecastMetrics`
  (per-(run,epoch) JSONL row), `StallTimer` (data-wait vs compute split), `MetricsLog`
  (prints, patches final val RMSE onto the last row, flushes JSONL).
- **`train_torch.py`** — loop instrumented; `--ceiling` trains an identical model on a
  RAM-preloaded epoch (no fetch/decode) = the compute-only in-memory ceiling.
- **`data.py`** — `--ceiling` / `--metrics-out` / `--size` / `--sample-chunk` /
  `--regenerate`; `make_advection_store` now **streams to bounded RAM** (chunk-aligned
  slab writes + raised `async.concurrency`) so `--source synthetic --url gs://…` writes a
  multi-GB cloud store idempotently (generate once, reuse on matching geometry).

Headline metric = **data-stall fraction** (share of wall-clock the GPU idles on the
loader; ~0 ⇒ compute-bound ⇒ loader effectively free) — self-referential, no competitor
baseline. `--ceiling` gives the fair substitute baseline.

Raw JSONL lives in `bench/results/` (gitignored); numbers summarized below.

### Findings

**WB2 ERA5 (128×64), read live from GCS, cuda** — the headline result:

| | steady samp/s | stall | % of ceiling | host RssAnon | val (model / persist) |
|---|---|---|---|---|---|
| insitu (no cache) | 1294 | 7.3% | **91.8%** | 1168 MB | 1.98 / 2.23 |
| insitu (NVMe cache) | 1294 | 6.2% | 92.1% | 848 MB | 1.99 / 2.23 |
| in-memory ceiling | 1410 | 0% | — | 1601 MB | — |

insitu keeps the GPU ~93% fed streaming ERA5 from `gs://`, at ~92% of the in-memory
ceiling, on **bounded** host memory (1168 MB vs the ceiling's 1601 MB full-epoch preload).
Cold epoch 0 is 24% stall / 885 ms TTFB, then warm epochs collapse to 4–12% as prefetch
fills.

**The NVMe cache is redundant here** — identical throughput ± noise. The no-cache run is
already compute-bound (7% stall), so there's nothing to remove; the cache only shifts
decoded chunks from heap (`RssAnon`) to reclaimable mmap (`RssFile`), lowering heap
1153→848 MB. Disk cache earns its keep only when the working set exceeds host RAM.

**Synthetic (48×48), cuda (GPU torture test — trivially cheap compute):** 5627 samp/s,
7.7% stall, 83.8% of ceiling, val 0.57 / 0.88. Floor case: near-zero compute leaves little
to hide IO behind, yet still 84% of ceiling.

**Large synthetic (256×256, ~13 GB on GCS), cuda (partial — 3 epochs before the box
hung):** stall **0.2–0.9%**, 144 samp/s, 149 MB/s, `RssAnon` ~14 GB, gpu 2207 MB. Fully
compute-bound: at 256² the CNN is ~222 ms/batch, so the loader is invisible even cold.

Two takeaways from the large run:
1. **Growing the field can't reach an IO-bound regime.** For this conv model, MB/s demand
   is ~size-invariant — bytes/sample and compute/sample both scale with pixel count and
   cancel, pinning demand near ~150 MB/s regardless of `--size`. Exposing the loader
   requires choking *supply* (low `--max-inflight` / GRIB-tiny `--sample-chunk`), not a
   bigger field.
2. The 14 GB `RssAnon` is the in-memory decoded cache holding the **entire** working set —
   which is why epoch-1+ stall is ~0 (served from RAM). It's currently **unbounded** (OOM
   risk on smaller boxes); an in-memory cache-budget knob would let us force eviction and
   make the NVMe-cache comparison bite without a store larger than host RAM.

### Verification
- CPU-validated end to end (timing, JSONL schema, val-RMSE patching); `--source synthetic
  --url gs://…` written to a real bucket, read back, trained (model beats persistence),
  cleaned up.
- `ruff` / `mypy` clean on `src bench examples`; `pytest -q` 143 passed / 11 skipped
  (`test_advection` green — streamed generator preserves the learnable signal).

### Bottom line
Across WB2 and a 13 GB synthetic store, **insitu is compute-bound with the loader
effectively free** — the strong result. Not merged-blocking, but the remaining round-1
datapoint is one *throttled* run (`--max-inflight 2`) to characterize the failure edge;
round 2 (pynvml per-step, H2D split) and the JAX/TF port are separate follow-ups.

### DIFF
--- a/examples/_forecast_metrics.py
+++ b/examples/_forecast_metrics.py
@@ -1,0 +1,200 @@
+"""Framework-neutral training metrics for the forecast examples.
+
+The headline is the **data-stall fraction**: the share of wall-clock the compute
+device idles waiting on the loader. Prefetch hiding IO behind compute drives it
+toward zero -- the run is compute-bound and the loader is effectively free. It is
+self-referential (no competitor baseline), so it replaces an xbatcher A/B.
+
+The paired ``--ceiling`` run feeds the *same* model + loop from a RAM-preloaded
+list of batches (no fetch, no decode); its throughput is the compute-only ceiling
+and ``insitu_samples_per_s / ceiling_samples_per_s`` is the loader overhead.
+
+This module owns everything framework-independent: the per-(run, epoch) JSONL
+schema (:class:`ForecastMetrics`), the stall/throughput timer, and the in-memory
+log that patches the final validation RMSE onto the last row before flushing. The
+torch loop supplies only the two framework-specific numbers (a synced step and the
+GPU peak); a later JAX/TF port reuses the same collector with their own sync.
+"""
+
+from __future__ import annotations
+
+import json
+import math
+import platform
+import socket
+import time
+from collections.abc import Iterable, Iterator
+from dataclasses import asdict, dataclass, field
+from pathlib import Path
+
+# Reuse the bench suite's RssAnon reader so host-memory is measured the same way the
+# storage benchmarks report it (heap/stack only -- the true bound; mmap cache pages
+# are reclaimable and excluded). bench is a repo-root dev package; the examples run
+# from the repo root (``python -m examples...``), so this import resolves there.
+from bench.result import rss_breakdown_mb
+from insitubatch.types import Batch
+
+
+@dataclass
+class ForecastMetrics:
+    """One self-describing row per (run, epoch) for the forecast training benchmark.
+
+    ``run`` is ``insitu`` (fed by the loader) or ``ceiling`` (fed from RAM). The
+    headline is ``data_stall_fraction``; ``val_*`` are filled only on the final row
+    of a run (one training run's forecast skill, a sanity check that it really trained).
+    """
+
+    run: str  # insitu | ceiling
+    framework: str  # torch | jax | tf
+    source: str  # synthetic | wb2 | arraylake
+    device: str  # cpu | cuda
+    epoch: int
+    n_samples: int
+    n_batches: int
+    batch_size: int
+    wall_s: float
+    data_wait_s: float
+    compute_s: float
+    data_stall_fraction: float  # data_wait_s / wall_s -- the HEADLINE
+    samples_per_s: float
+    mb_per_s: float  # decoded volume delivered to the model
+    mean_data_wait_ms: float  # per batch
+    mean_compute_ms: float  # per batch
+    ttfb_ms: float  # wait for the first batch of the epoch (cold start on epoch 0)
+    peak_rss_anon_mb: float  # heap/stack high-water at epoch end (the real host bound)
+    peak_gpu_mem_mb: float  # torch.cuda peak this epoch (0 on cpu)
+    val_model_rmse: float = math.nan  # set on the last row of a run only
+    val_persistence_rmse: float = math.nan
+    host: str = field(default_factory=socket.gethostname)
+    platform: str = field(default_factory=platform.platform)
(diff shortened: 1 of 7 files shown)


examples/_forecast_metrics.py
```diff
@@ -0,0 +1,200 @@
...
+        t = time.perf_counter()
+        for i, batch in enumerate(source):
+            now = time.perf_counter()
+            wait = now - t
+            self.data_wait += wait
+            if i == 0:
+                self.ttfb = wait
+            self.n_batches += 1
+            self.n_samples += len(batch.sample_indices)
+            self.total_bytes += sum(a.nbytes for a in batch.arrays.values())
```
emfdavid, line 101: This should be size * len?

examples/advection/train_torch.py
```diff
@@ -58,32 +61,120 @@ def _forecast(model: AdvectionCNN, batch: Batch, device: torch.device) -> torch.
...
+    return model
+
+
+def _val_rmse(model: AdvectionCNN, ds: InSituDataset, dev: torch.device) -> tuple[float, float]:
     model.eval()
     with torch.no_grad():
         return evaluate(ds.val, lambda b: _forecast(model, b, dev).detach().cpu().numpy())
 
 
+def train(
```
emfdavid: Can we separate this into a train_torch_metrics module so we don't complicate the usage example?

emfdavid (commented):
Nice metrics and persistent dataset - the train_torch module is a canonical example - should not be confused with the metrics work.

emfdavid:
Addressed both review points in 606eb20:

**Split the metrics out of the canonical example.** `train_torch.py` is back to the plain train loop (nothing but the model + step + eval); the stall/ceiling instrumentation moved to a new **`train_torch_metrics.py`** that imports the same `AdvectionCNN`/`_forecast` (no duplication). The `--ceiling`/`--metrics-out` flags now live only on the metrics tool — `build_parser()` in `data.py` is the shared canonical CLI, and the metrics module extends it — so they no longer show up on the canonical example's `--help`. New invocation: `python -m examples.advection.train_torch_metrics ...`. Added a `test_advection` case covering the collector (stall/ceiling rows, wall = wait + compute, val-RMSE patched onto the final insitu row, JSONL mirror).

**`sum(a.nbytes ...)` timing.** `batch.arrays.values()` is the ~4 field arrays, not grid cells, and `nbytes` reads shape×itemsize off the header (no data scan). Best-of-five it's flat at ~0.57 µs from a 2 MB to a 268 MB batch — constant-time and cheaper than a `size*len` computation — so against a 5–222 ms step it's ~1e-4 %. Left as-is.

Unrelated: the `test-freethreaded` red is a transient crates.io download failure (`[16] Error in the HTTP2 framing layer`) building mypy's `ast-serialize` Rust dep from source — it also failed on the latest `main` run, so it's pre-existing CI flakiness, not this PR.

PR #17, opened 2026-07-15
examples: SDSS spectral reconstruction (spPlate, train in place, no reshard)

emfdavid:
An astroML cross-domain example: reconstruct SDSS galaxy spectra streamed in place from real
`spPlate` FITS, with no reshard. The decode-amortization/streaming companion to the Hubble example.

## What it does

astroML's spectral-PCA workflow (`fetch_sdss_corrected_spectra` → `compute_sdss_pca`) first
downloads the raw archive and resamples every spectrum onto a common grid into one `spec4000.npz` —
the download-and-reshard step insitubatch argues against. `examples/sdss` instead indexes real
`spPlate` frames on `data.sdss.org` (HTTPS) as **virtual references**: one plate holds all ~640
fibers on a common log-wavelength grid as a single `(fiber, wave)` image, which is virtually
re-chunked along the fiber axis by **pure byte arithmetic** — no pixels moved.

The task mirrors astroML's spectral reconstruction: a 1-D **convolutional autoencoder**, trained by
SGD over the streamed mini-batches, beats a **PCA baseline** at the same latent dim — because galaxy
spectra are translation-structured (varying redshift shifts the lines, which a fixed-dim linear
basis reconstructs poorly). The offline synthetic default (redshift-shifted spectra) carries the
beats-PCA claim and backs the drift test; the real run is a same-pipeline, real-archive demo.

## Two build modes (`--plates N`) — both rewrite only the chunk manifest, no pixels moved

- **one plate** → full wave width, many fibers per chunk (the O(chunks) *decode-amortization* regime).
- **many plates** → every SDSS plate shares `dloglam` and their `COEFF0` start wavelengths differ by
  whole bins, so cropping each to the shared window aligns them onto one grid *exactly* (no
  resampling) and folds `(plate, fiber)` into one flat sample axis, one fiber per chunk (the
  *streaming* regime — the only no-reshard way to concat the ragged-width plates, scaling to the
  full ~2800-plate archive). Verified: 3 plates → 1920 spectra byte-identical to astropy.

The modes are mutually exclusive by construction: cropping to the common window breaks the
multi-fiber contiguity the single-plate mode rides on — an honest boundary of the FITS byte layout.

## FITS binary tables — upstream fixes now released

SDSS spectra also ship as FITS binary tables (structured big-endian dtypes).
`tests/test_fits_bintable.py` validates that path end-to-end through insitubatch (kerchunk →
VirtualiZarr → Icechunk → insitubatch) against astropy.

The three upstream fixes this path needed are all shipped in
[VirtualiZarr 2.7.2](https://github.com/zarr-developers/VirtualiZarr/releases/tag/v2.7.2), so the
`[tool.uv.sources]` git pin is gone — the `astronomy` extra is plain released wheels:

- [VZ#1037](https://github.com/zarr-developers/VirtualiZarr/pull/1037) — the kerchunk translator
  reads the list-valued struct field specs (released VirtualiZarr used to raise).
- [VZ#1047](https://github.com/zarr-developers/VirtualiZarr/pull/1047) +
  [VZ#1048](https://github.com/zarr-developers/VirtualiZarr/pull/1048) — the bytes codec records
  the struct's big-endian byte order (and reads it across zarr's `Endian` enum → `str` change).

With the byte order recorded, zarr's half engages —
[zarr-python#4142](https://github.com/zarr-developers/zarr-python/pull/4142), first released in
**3.3.0** — and the fields arrive already swapped. So the `.view('>f4')` reinterpret this PR
originally documented at the numpy boundary is now *wrong*, not merely redundant: it would swap
back to garbage. The tripwire test written to catch exactly that fired on the upgrade and is
inverted here into a guard that the delivered field is native-order **and** holds the true values.
The extra floors `zarr>=3.3` so an older zarr cannot silently deliver plausible garbage instead.

## Tests / gates

`tests/test_sdss.py` (synthetic drift guard, AE beats PCA), `tests/test_sdss_build.py` (both build
modes offline, exact cross-plate alignment, fail-fast on non-bin-aligned grids), and
`tests/test_fits_bintable.py`. Rebased onto current `main`; full suite 285 passed / 18 skipped on
zarr 3.3.0 + VZ 2.7.2; ruff + format + mypy clean.

Because the extra no longer needs a git pin, **CI now installs it**: a new
`--extra torch --extra astronomy` test leg runs the FITS tests (offline, synthetic FITS in
`tmp_path` — no network) instead of importorskip-skipping them, and the lint job gains the extra so
mypy resolves the real VirtualiZarr types in `examples/`.

> Follow-up, deliberately not in this PR: `docs/examples.md` (the canonical examples page, added on
> main after this branch) still owes an SDSS row + section, per the plan to do the docs pass once
> after merge. SDSS is HTTPS-only (no S3 mirror), so throughput benchmarking co-location is still
> TBD.


martindurant:
Since you pick a concrete example with real data, this is clearly what at least some astronomers do :) 
My experience is pretty outdated here, and I was only ever looking at individual objects anyway.

Is there an intermediate representation of the chunk/manifest that can be persisted in a public place? If the workflow were structured as 1) scan 2) use, then interested people could just do the latter with any benchmarking they care to do. I understand its not quite just a vzarr manifest if you need to do extra work like numpy  .view() calls.

Will you make this a published blog or something?

emfdavid:
Good question, and it's the right shape — scan-once / use-many is already how the example is structured, and Icechunk is the persistence layer that makes the "scan" output publishable.

The one thing I want to keep central is that this stays anchored to the tools astronomers already validate with. **astropy is the ground truth here**: the build checks byte-identical against `astropy.io.fits` (3 plates → 1920 spectra, exact), and that comparison — not the manifest — is how you know a streamed batch is correct. A persisted Icechunk manifest accelerates the *use* step; it never replaces astropy as the reference. (The `.view('>f4')` you flagged is part of *staying* byte-exact to astropy at the numpy boundary — a shim, not a divergence. [VZ #1047](https://github.com/zarr-developers/VirtualiZarr/pull/1047) + [#1048](https://github.com/zarr-developers/VirtualiZarr/pull/1048), both mergeable, remove the need for it without changing the astropy-validated result — and it only affects the FITS *binary-table* path; the spPlate image path is already clean.)

On persistence itself: the committed Icechunk store isn't an extra step — it's a **one-time side effect of the first build** (kerchunk → VirtualiZarr → Icechunk → insitubatch). The scan writes it once — virtual refs, no pixels moved — and everything downstream already reads from it. So "publish the manifest" just means dropping that artifact in a public bucket: a consumer opens it and trains/benchmarks with no re-scan and no astropy in the loop. The raw FITS is HTTPS-only (no S3 mirror), but the manifest is tiny and location-independent.

Where it's headed makes the "what for" concrete: the eventual target is a **PR to astroML** adding insitubatch recipes that hook into its existing SDSS workflow (`fetch_sdss_corrected_spectra` / `compute_sdss_pca`) — train-in-place as an option *inside* the ecosystem astronomers already use, validated against the same astropy/astroML baselines, rather than a stack off to the side.

And yes — blog post committed, kicking off with a Pangeo Showcase talk (~September), with this SDSS case as one of the cross-domain examples.

emfdavid:
@martindurant — the scan-once / use-many split you asked for is live. The manifests are
published in a public bucket, read-anonymous, no requester-pays:

```
gs://insitubatch-bench-insitubatch/astronomy/
```

Three stores, all exposing one `flux (fiber, wavelength)` float32 array:

| store | shape | chunks | fibers/chunk |
|---|---|---|---|
| `sdss_dr17_p0266_refs` | 640 × 3864 | 64 × 3864 | 64 |
| `sdss_dr17_6plate_refs` | 3840 × 3841 | 1 × 3841 | 1 |
| `sdss_dr17_6plate_mirror_refs` | 3840 × 3841 | 1 × 3841 | 1 |

Step 2 needs `icechunk` and nothing else — no kerchunk, no VirtualiZarr, no astropy, no
credentials:

```python
import icechunk, zarr

BUCKET = "insitubatch-bench-insitubatch"
PREFIX = "astronomy/sdss_dr17_6plate_mirror_refs"
PIXELS = f"https://storage.googleapis.com/{BUCKET}/astronomy/sdss/dr17/"

config = icechunk.RepositoryConfig.default()
config.set_virtual_chunk_container(
    icechunk.VirtualChunkContainer(PIXELS, icechunk.http_store())
)
repo = icechunk.Repository.open(
    icechunk.gcs_storage(bucket=BUCKET, prefix=PREFIX, anonymous=True),
    config=config,
    authorize_virtual_chunk_access=icechunk.containers_credentials({PIXELS: None}),
)
flux = zarr.open_array(repo.readonly_session("main").store, path="flux", mode="r")
flux.shape   # (3840, 3841)
```

**And the `.view('>f4')` you flagged is gone.** That is entirely down to upstream: VirtualiZarr
[#1037](https://github.com/zarr-developers/VirtualiZarr/pull/1037) (list-valued kerchunk field
specs), [#1047](https://github.com/zarr-developers/VirtualiZarr/pull/1047) (detect a struct's
big-endian byte order and record it in the bytes codec) and
[#1048](https://github.com/zarr-developers/VirtualiZarr/pull/1048) (the `Endian` enum→str
change), shipped in **2.7.2**, plus zarr-python
[#4142](https://github.com/zarr-developers/zarr-python/pull/4142) in **3.3** doing the decode
half. With those released the git pins are dropped, the reinterpret shim is deleted, and the
delivered fields are native-order and byte-identical to astropy with nothing in between. Thanks
to the VirtualiZarr and zarr-python maintainers for turning those around quickly — and to you,
since `kerchunk.fits` is what parses the headers underneath all of it.

Two things worth knowing before anyone benchmarks against these:

**The layouts are a real trade-off, not a tuning knob.** `p0266` is one plate re-chunked along
the fiber axis at full width — 64 spectra per read, so decode amortizes. The 6-plate stores
crop every plate to the shared log-wavelength window so they concatenate onto one grid exactly
(same `dloglam`, starts differing by whole bins, no resampling) — and that crop breaks the fiber
contiguity, leaving one fiber per chunk. Archive scale and decode amortization are mutually
exclusive here; it falls out of the FITS byte layout.

**One fiber per chunk is latency-bound**, so read concurrency is the only knob that matters. Full
pass over all 3840 spectra from a VM in us-central1, medians of five interleaved runs:

| reader | concurrency | full pass | spectra/s |
|---|---|---|---|
| zarr, 64-row slices | 10 (default) | 22.3 s | 172 |
| zarr, 64-row slices | 32 | 12.6 s | 305 |
| insitubatch | 32 | 10.7 s | 360 |

Both readers knee at ~32 and neither improves past it. Note the first two rows: zarr's
`async.concurrency` default of 10 is worth raising for a store shaped like this — comparing
against the default rather than a tuned baseline would have let me claim about twice the margin
that is actually there. The honest number is ~18%, and it is small for a good reason: one sample
per chunk is precisely the geometry where a batch loader has least to offer, since there is
nothing to amortize within a chunk and no redundant reads to de-duplicate. What it buys is
shuffled, split-aware, transformed batches rather than raw slices.

Bucket docs are in
[`astronomy/README.md`](https://storage.googleapis.com/insitubatch-bench-insitubatch/astronomy/README.md)
(geometry, wavelength solution, plain-zarr recipe) and
[`PROVENANCE.md`](https://storage.googleapis.com/insitubatch-bench-insitubatch/astronomy/sdss/dr17/PROVENANCE.md)
(sources, SDSS public-domain statement, citation). Longer write-up with the same material:
https://emfdavid.github.io/insitubatch/sdss-dataset/

astropy remains the ground truth — the build still checks byte-identical against
`astropy.io.fits`, and the published stores are that same artifact. Blog post and the Pangeo
Showcase talk (~September) still on; this will be one of the cross-domain cases.

emfdavid:
Correction to the throughput table in my comment above — and a caveat on how much weight any
of it should carry.

The numbers were measured with a connection-pool configuration that none of the published
recipes use, so they were not reproducible from the code I linked. The benchmark passed
`icechunk.http_store({"connect_timeout": "30s", "pool_idle_timeout": "300s",
"pool_max_idle_per_host": "4"})`, carried over from an unrelated fix for `data.sdss.org`
dropping connections; every recipe in the docs and the bucket READMEs uses a bare
`icechunk.http_store()`. Re-measured with the bare store, same interleaved method, five runs
each, from a VM in us-central1:

| reader | concurrency | full pass (3840 spectra) | spectra/s |
|---|---|---|---|
| zarr, 64-row slices | 10 (default) | 17.9 s | 215 |
| zarr, 64-row slices | 32 | 12.2 s | 314 |
| insitubatch | 32 | 7.3 s | 530 |

The mechanism is worth stating, because the error ran in my favour and I would rather explain
it than quietly swap a table. `pool_max_idle_per_host: 4` caps idle keepalive connections well
below the read concurrency of 32, which penalises a *sustained* reader far more than a bursty
one: insitubatch keeps 32 requests in flight for a whole epoch and went 10.7 s → 7.3 s when the
cap came off, while zarr's slice loop — burst, assemble, burst — barely moved, 12.6 s → 12.2 s.
zarr holding steady across both sessions is the control that says this is the setting rather
than network drift.

**I am not claiming a speed win on the strength of this, and I would rather it not be read as
one.** All I have tuned on the zarr side is `async.concurrency`; I have not swept slice width,
alternative access patterns, thread-pool settings, or a different store backend, and any of
those could narrow or close the gap. Two reasonable configurations measured fairly against each
other is not the same as a tuned-versus-tuned result, and I have already been wrong once in
this thread by comparing against a default. If someone wants to tune the zarr arm properly, the
stores are public and I will happily publish whatever comes back.

Parity is the bar that matters to me here anyway. At one fiber per chunk there is nothing to
amortise within a chunk and no redundant reads to de-duplicate, so a batch loader has no
structural advantage to convert — reading at roughly the speed of a straightforward zarr read
is the expected outcome, and it is sufficient. The argument for the loader on this geometry is
what you get *at* that speed: shuffled, split-aware, chunk-aligned batches with transforms,
streamed with no reshard and flat memory, instead of raw slices you would then have to build
all of that on top of.

Where chunk layout is the variable, the effect is much larger and much less ambiguous, because
it is the same reader against itself. Plate 0266 built both ways from the same mirrored bytes,
same 640 spectra, verified byte-identical:

| reader | layout | full pass (640 spectra) | spectra/s |
|---|---|---|---|
| insitubatch | 64 fibers/chunk | 0.11 s | 5,835 |
| insitubatch | 1 fiber/chunk | 1.26 s | 507 |
| raw zarr | 64 fibers/chunk | 0.43 s | 1,484 |
| raw zarr | 1 fiber/chunk | 1.65 s | 388 |

**11.5× from the chunk geometry alone** for insitubatch, 3.8× for raw zarr over the identical
change. Both readers benefit; the point I would stand behind is that the layout dominates, not
that one reader beats the other. (These passes are short enough that cold start dominates the
spread, so treat the ratio as the durable part and the absolute times as indicative.)

One other change since that comment: the three published stores now carry `COEFF0`/`COEFF1` as
array attributes, so the wavelength axis is reconstructable from the store rather than from a
table in a README. It went in as a metadata-only snapshot, so paths and history are unchanged;
verified anonymously against the source FITS headers, with the pixels re-checked byte-identical
to astropy.

The docs page and the bucket READMEs carry the corrected numbers and the same caveat.

PR #21, opened 2026-07-31
Reuse batch output buffers; opt-in page-locked H2D path

emfdavid:
Reuse the batch output buffers instead of allocating one per batch, and add an opt-in
page-locked H2D path on top of that pool.

## What it's worth — a range, not a headline

Both changes remove **loader-side** cost, so their end-to-end value is bounded above by an
isolated probe with that cost fully exposed, and below by **zero** — a loop whose compute step
already hides the loader collects nothing. That's prefetch working, not a disappointment. The
useful output is the bounds plus a rule for locating yourself between them.

| | upper bound (loader cost exposed) | this GPU training loop | lower bound |
|---|---|---|---|
| **reuse** | +10 … +16% above 32 MiB | **≈ 0** (±0.3 pt) | 0 |
| **pinning** | copy hidden entirely below ~37 MiB | **+1 … +2%** | 0 |

The asymmetry is structural, not luck. **Reuse saves producer-thread time — exactly what
prefetch is built to hide. Pinning shortens a copy on the consumer's critical path, which
prefetch cannot touch.** So a saturated training GPU is reuse's *floor*, not an interior point,
and the top of these ranges is collected by workloads spending less compute per byte:
inference over training, shallow models, cold caches, IO-bound epochs.

Reuse's floor value isn't a speedup at all — it's the guarantee that growing a batch past
glibc's 32 MiB `mmap` threshold cannot walk you off a cliff.

### Reuse, measured across the threshold

Batch size swept over one 256² store, no-reuse arm from a checkout identical but for the
`gather` call site, each arm scored against **its own** in-session ceiling:

| MiB per variable buffer | 8 | 16 | 32 | 64 |
|---|--:|--:|--:|--:|
| reuse (% of ceiling) | 99.69 | 99.92 | 99.74 | 98.74 |
| no reuse | 99.49 | 99.72 | 99.92 | 99.03 |
| delta (points) | +0.21 | +0.20 | −0.18 | −0.29 |

Inside ±0.3 with the sign flipping twice, across an 8× range spanning the threshold. At 64 MiB
the two arms agree to **0.04%** absolute (138.52 vs 138.57 samples/s). The loop runs at
98.7–100% of compute ceiling, so allocation has nowhere to surface *even where the `mmap` path
demonstrably engages* (0 faults/alloc at 8 and 16 MiB, 100% of pages at 32 and 64).

### Pinning, bracketed against drift

**+1 to +2%.** Don't quote a third digit: an A/B/A on an unchanged checkout read
145.4 / 146.6 / 143.4 samples/s, so two *identical* unpinned blocks differed 1.4% twenty
minutes apart — drift is ~0.7% per block, half the signal. Linear fit gives +1.5%; the bounds
are +0.8% (`b−a`) to +2.2% (`b−c`); cross-session ceilings said +2.0–2.3%.

**Method, since it cost us a night:** run the control arm *twice*, bracketing the treatment,
and interpolate. And `% of ceiling` isolates reuse but **cancels** pinning — `--pin` reaches
the ceiling too, because `pin_host_buffers` precedes `preload_epoch` (= `list(ds.train)`), so
the ceiling moves in lockstep.

## Three defects found and fixed

1. **Uncalibrated liveness baseline** — the idle refcount was recorded while the allocation's
   own reference was still live, so every buffer looked permanently free and the pool lent
   **one allocation to every batch at once**, each overwriting the last. Pinned path only,
   because `torch.empty(pin_memory=True).numpy()` returns an array that *is* its own data
   owner while the aligned-heap default returns one that *views* an owner. Confirmed on an L4:
   `double-lent: True | first batch: 2.0` before, `False | 1.0` after. Its fingerprint —
   a refcount reading *below* the calibrated idle baseline — now raises, so this class cannot
   come back silently (see the review round).
2. **Non-atomic check-then-lend** — two producers (`zip(ds.train, ds.val)`, two DataLoaders)
   could both observe one buffer free and write into it. Now one locked step.
3. **`_InFlight.hold` unlocked read-modify-write** — prune-then-append on process-wide state;
   a thread could overwrite a hold another appended and free a buffer mid-DMA. Window is a few
   bytecodes and does **not** reproduce unaided (the comprehension iterates by index, so it
   picks up concurrent appends); widened at exactly that point it loses a live hold 20/20.

## The thing that actually caught bug 1

Neither throughput nor forecast skill did. **Persistence RMSE is model-independent** —
`rmse(t2m(t), t2m(t+24h))`, no training, no batch size — so it's a fingerprint of the bytes the
loader delivered, bit-identical at `0.588030696` across every batch size, arm, and repeat.

- Throughput missed it: the same bytes still moved.
- Skill missed it: the corrupt arm scored **+13.6%**, *above* real ERA5's honest +11.4%.
- The old summary line hid it: it printed `persistence/model` as "Nx better", so a model **84%
  worse** than predicting no change rendered as `(0.5x better)`.

Now: signed skill score with the verdict on **sign alone** (no threshold is sound — healthy
skill is +33/+22/+10% on synthetic grids and +11% on real ERA5, and a corrupt run beat ERA5),
persistence printed as a labeled invariant, and `PersistenceCheck` in `advection_sweep`
asserting it per store across repeats, configs and arms — loud inline, non-zero at exit.

## Also in here

- **`INSITUBATCH_NO_BUFFER_REUSE=1`** — escape hatch and A/B control. Reuse is safe against
  everything the transform API can do (a transform returning a derived array frees its buffer
  early, one returning a view stays tracked, one that retains makes the pool allocate
  elsewhere — all read correct data), so this is not a tuning knob; it covers memory escaping
  Python object semantics, and being wrong. It also retires the worktree the no-reuse arm
  needed. Epoch log stamps `REUSE OFF`.
- **Per-epoch INFO line** — chunk-cache hits, resident peak, and pool state. `allocated`
  staying nonzero after epoch 0 is the signal that batches aren't coming back.
- **`--log-level` on the examples**, raising only the `insitubatch` logger.
- **`_torch_dtype`** — numpy→torch mapping from torch's own converter rather than
  `getattr(torch, str(dtype))`. Not a bug fix; removes a second mechanism answering a question
  `to_torch` already answered, so buffer and tensor dtype cannot diverge.
- **Docs**: `DESIGN.md` described weakref liveness, which cannot work (numpy collapses base
  chains, so a crop doesn't reference the view it came from) and was never built; its reuse
  null was measured at 2 MiB, below the cliff; its pinning table came from voided runs.
  `tuning.md` now documents the batch queue as a high-water mark.

## Considered and rejected

- **A `reuse=False` parameter.** Its documentation would have to teach numpy base-chain
  collapsing and `sys.getrefcount`. The env var is a kill switch with a one-sentence doc, not
  a knob asking users to audit our invariant.
- **A public `trim()`.** The pool's floor can never exceed a peak the process already survived
  — it *is* what was simultaneously live at that peak, which fresh allocation also held. API
  surface for a problem with no demonstrated workload.
- **Refusing pinning + `INSITUBATCH_NO_BUFFER_REUSE`** (review). It warns instead — see below.

## Review round

Three comments, three changes.

- **`refs < idle_refs` is now an error.** Every reference counted at calibration is structural
  (`self.owner`, the base chain `self.base` holds), so all of them outlive the buffer and the
  count can only rise. A lower reading is not a near-miss — it means the baseline swallowed a
  transient, which makes `free` permanently true and lends one allocation to every batch at
  once. That is bug 1's exact fingerprint, and nothing raised on it; it had to be found by
  persistence RMSE. One integer compare against a path copying megabytes, so checking forever
  is free. Test reproduces it the way the original arose (calibrate with an extra owner
  reference live — what `torch.empty(pin_memory=True).numpy()` did by leaving the return value
  on the caller's stack) plus the converse, that ordinary lend/derive/reclaim never trips it.
- **Pinning with reuse off warns, and stays reachable.** It deserves to be loud for a reason
  the docstring didn't have: `pinned_allocator`'s budget is a monotone counter sized for a pool
  that converges, so without reuse it's spent within a few batches and the rest of the run
  finishes pageable while `kind` still reports pinned intent — misleading as a timing row, not
  merely slow. `benchmarks.md` now says so. But erroring would remove the one control that
  isolates reuse from pinning, and the double-lend lived on the pinned path and nowhere else:
  a user suspecting corruption would have to change the allocator too, which is two variables
  and answers nothing. The check sits in `set_allocator` (the single injection point), so it
  covers `as_torch` and a user-supplied labelled allocator alike; a parametrized test asserts
  the other three combinations stay silent.
- **`probe_batch_buffers` is pinned to the core API — it had already drifted.** The probe
  predates the implementation and carried its own `aligned_empty` and `XLA_ALIGN` plus
  `getattr(torch, str(dtype))` in three places, the very mechanism `_torch_dtype` was added to
  remove. A change in `src/` would have left it reporting the old design's numbers, visible
  only on a GPU box. Now imports all three; both swaps are behaviour-identical, so no published
  number moves. Tests assert *identity*, not equality — an equal-looking reimplementation is
  the drift being prevented — and run what CI can reach: `main()` end to end on the host arm
  over a toy case, unknown-arm rejection, every `CASES` dtype through the pool's allocator and
  torch's converter, and the jax arm asserting its two device checks report **unknown** without
  a GPU rather than *safe*. `Roundtrip.verdict` and `JaxCheck.poll_insufficient` get
  table-driven tests: pure logic, no hardware needed, previously uncovered.

## Open, not blocking

Whether `sys.getrefcount` is safe under free-threading. The 3.13t CI job runs the full suite
GIL-off, but only one staleness direction is dangerous (reading *low* → reuse a live buffer)
and that direction is unproven. The new baseline guard changes what a stale-low read *does* —
it raises where it previously corrupted silently, which is the direction we want, but it means
3.13t can now go red where it was quietly wrong.

## Verification

266 passed, 12 skipped; `ruff` and `mypy` clean on `src bench examples`.

### DIFF
--- a/src/insitubatch/buffers.py
+++ b/src/insitubatch/buffers.py
@@ -39,7 +39,7 @@ pinned path, since page-locked memory is page-aligned.
 
 from __future__ import annotations
 
-import weakref
+import sys
 from collections.abc import Callable
 
 import numpy as np
@@ -66,24 +66,40 @@ def aligned_empty(shape: tuple[int, ...], dtype: np.dtype, align: int = XLA_ALIG
 
 
 class _Buffer:
-    """One owned base array plus a weak handle on the view most recently lent from it."""
+    """One allocation, tracked by how many arrays still reference its data owner."""
 
-    __slots__ = ("base", "lent")
+    __slots__ = ("base", "owner", "_idle_refs")
 
     def __init__(self, base: np.ndarray) -> None:
         self.base = base
-        self.lent: weakref.ref[np.ndarray] | None = None
+        # Every array viewing this memory -- the batch we lend, a transform's crop of it, a
+        # torch/JAX tensor's hold on either -- keeps a strong reference to the object that
+        # *owns* the data. numpy collapses base chains to that owner rather than chaining, so
+        # a lent view is NOT referenced by a slice taken from it; the owner is. Probing with a
+        # throwaway slice is what identifies it, since it differs by allocator (heap arrays
+        # own their own data; a page-locked one is a view of a torch tensor's buffer).
+        probe = base[:1]
+        self.owner = base if probe.base is None else probe.base
+        del probe  # the probe is itself a reference; the idle baseline must not include it
+        self._idle_refs = 0
+        self._idle_refs = self._refs()
+
+    def _refs(self) -> int:
+        """References to the data owner. Must be reached through exactly one call frame so
+        the idle baseline and the live count are measured identically."""
+        return sys.getrefcount(self.owner)
 
     @property
     def free(self) -> bool:
-        """True when nobody holds the last view we lent -- so the memory is ours to rewrite.
+        """True when nothing outside this pool references the memory, so it is ours to rewrite.
 
-        ``lent() is None`` is the whole safety argument: CPython drops a view on its last
-        decref, and every zero-copy export we support (``torch.from_dlpack``,
-        ``jnp.from_dlpack``) takes a strong reference to that view, so a live tensor keeps
-        this False.
+        Counting references to the *owner* rather than weak-referencing the lent array is what
+        makes this survive a derived view: CPython drops each array on its last decref, and
+        every zero-copy export we support (``torch.from_dlpack``, ``jnp.from_dlpack``) holds
+        the array it was given, which in turn holds the owner. A crop the consumer kept after
+        discarding the batch therefore still counts.
         """
-        return self.lent is None or self.lent() is None
+        return self._refs() <= self._idle_refs
 
 
 class BatchBuffers:
@@ -114,15 +130,13 @@ class BatchBuffers:
 
     @staticmethod
     def _lend(buf: _Buffer, n: int) -> np.ndarray:
-        """Hand out a fresh view of ``buf`` and record a weak handle on it.
+        """Hand out a view of ``buf``'s first ``n`` rows.
 
-        The view must be a new object each time -- that is what the weakref tracks. Slicing
-        always builds one, and slicing the full length is what lets a ragged batch and a full
-        batch share the same base.
+        Slicing is what lets a ragged final batch and a full one share an allocation; the
+        returned array references the buffer's owner, which is how :attr:`_Buffer.free`
+        sees it.
         """
-        view = buf.base[:n]
-        buf.lent = weakref.ref(view)
-        return view
+        return buf.base[:n]
 
     @property
     def nbytes(self) -> int:
--- a/src/insitubatch/buffers.py
+++ b/src/insitubatch/buffers.py
@@ -9,10 +9,17 @@ count tracks batches rather than dataset size. Paid per batch, that is a fresh p
 fault in plus ``munmap``'s TLB work.
 
 Below the threshold the freed block is recycled on the heap instead and reuse buys nothing --
-it costs about 2-3%. End to end (``bench/batch_buffer_sweep.py``) the pool is *flat* across a
-16x payload range while fresh allocation falls off a cliff at 32 MiB: 8 MiB -2.4%, 32 MiB
-+16%, 64 MiB +13%, 128 MiB +10%. So this removes a cliff rather than adding speed. Weather
-batches sit well below it; ViT and microscopy batches sit above.
+it costs about 2-3%. Allocation-bound (``bench/batch_buffer_sweep.py``) the pool is *flat*
+across a 16x payload range while fresh allocation falls off a cliff at 32 MiB: 8 MiB -2.4%,
+32 MiB +16%, 64 MiB +13%, 128 MiB +10%.
+
+**In a compute-bound loop it is worth nothing, and that is the expected result.** The GPU
+advection sweep A/B'd reuse against no-reuse at 8/16/32 MiB per buffer and came back inside
++-0.2 points of ceiling with the sign flipping -- because that loop runs at 99.5-100% of its
+compute ceiling, so producer-side allocation has nowhere to surface even at 32 MiB where the
+``mmap`` path demonstrably engages. So this removes a cliff for the loops that are *not*
+already compute-bound; it does not make a fed GPU faster. Weather batches sit below the
+threshold anyway; ViT and microscopy batches sit above it.
 
 **Hand-out is by view, reclaim is by liveness.** A buffer is lent as a view of a base array the
 pool owns forever, and comes back only once that view is unreferenced. The check happens
@@ -39,6 +46,7 @@ pinned path, since page-locked memory is page-aligned.
 
 from __future__ import annotations
 
+import os
 import sys
 import threading
 from collections.abc import Callable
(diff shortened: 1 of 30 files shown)


src/insitubatch/buffers.py
```diff
@@ -0,0 +1,353 @@
...
+    def free(self) -> bool:
+        """True when nothing outside this pool references the memory, so it is ours to rewrite.
+
+        Counting references to the *owner* rather than weak-referencing the lent array is what
+        makes this survive a derived view: CPython drops each array on its last decref, and
+        every zero-copy export we support (``torch.from_dlpack``, ``jnp.from_dlpack``) holds
+        the array it was given, which in turn holds the owner. A crop the consumer kept after
+        discarding the batch therefore still counts.
+        """
+        return self._refs() <= self._idle_refs
```
emfdavid: Should refs < idle_refs be an error?

src/insitubatch/buffers.py
```diff
@@ -0,0 +1,353 @@
...
+import threading
+from collections.abc import Callable
+from typing import NamedTuple
+
+import numpy as np
+
+XLA_ALIGN = 128
+"""Alignment XLA:CPU requires to import a DLPack buffer without copying."""
+
+NO_REUSE_ENV = "INSITUBATCH_NO_BUFFER_REUSE"
```
emfdavid, line 73: Should pinning with NO_REUSE_ENV be an error?

emfdavid (commented):
Need high level test coverage for the probe_batch_buffers to avoid drift with core api.

src/insitubatch/buffers.py
emfdavid: Yes — done in 8a5b71f.

It is not a near-miss, it is bug 1's fingerprint. Every reference counted at calibration is structural (`self.owner`, and the base chain `self.base` holds), so all of them outlive the buffer and the count can only rise. A reading *below* the baseline means the baseline itself swallowed a transient — which makes `free` permanently true, so the pool lends one allocation to every batch at once and each overwrites the last. Nothing raised on that; it had to be found by persistence RMSE.

One integer compare against a path that already copies megabytes, so checking forever is free. The only reachable ways to hit it are a bug in this file or a platform where `sys.getrefcount` does not mean what we assume, and both are better as a loud stop than as corrupted training data. `INSITUBATCH_NO_BUFFER_REUSE` is the way past it if it ever fires.

Test reproduces it the way the original arose — calibrate while an extra reference to the owner is live, which is exactly what `torch.empty(pin_memory=True).numpy()` did by leaving the allocator's return value on the caller's stack — then drop it. A second test asserts the guard stays quiet through ordinary lend / derive-a-view / reclaim cycles.

Worth flagging: under free-threading a stale-low read now raises where it previously corrupted silently. That is the direction we want, but it means the 3.13t job can go red where it was quietly wrong.

src/insitubatch/buffers.py
emfdavid: Warned, not refused — 8a5b71f.

Looking at it properly turned up a reason to be loud that the docstring did not have: `pinned_allocator`'s budget is a **monotone counter** sized for a pool that converges. Without reuse it is spent within a few batches and the rest of the run finishes on pageable memory, while `kind` still reports `pinned` intent. So the pairing is not merely slow — it is misleading as a *timing* row. `docs/benchmarks.md` now says that where it tells you to reproduce the no-reuse arm.

But it has to stay reachable. The double-lend lived on the pinned path and nowhere else, so refusing the pairing removes the one control that isolates reuse from pinning: a user who suspects reuse corruption would have to change the allocator too, which is two variables and answers nothing. That is the question the escape hatch exists for, and it is a correctness question, not a throughput one.

Check sits in `BatchBuffers.set_allocator` rather than `pin_host_buffers`, since that is the single injection point — so it covers `as_torch(device=...)` and a user-supplied labelled allocator alike. Parametrized test asserts the other three combinations stay silent: a warning on a supported configuration is a warning people learn to ignore.

emfdavid:
Re: high-level coverage for `probe_batch_buffers` — done in 3cb7177, and it had already drifted.

The probe predates the implementation, and carried its own `aligned_empty` and `XLA_ALIGN` plus `getattr(torch, str(case.dtype))` in three places — the exact second mechanism `_torch_dtype` was added on this branch to remove. A change to alignment or to the numpy→torch mapping in `src/` would have left the probe quietly reporting the old design's numbers, and only on a GPU box. It now imports all three from `insitubatch`. Both swaps are behaviour-identical, so no published measurement moves.

Coverage is deliberately **identity, not equality** (`probe.aligned_empty is core.aligned_empty`): asserting both return 128-byte-aligned arrays would pass against an equal-looking reimplementation, which is the drift being prevented.

The rest runs what CI can reach:

- `main()` end to end on the host arm over a toy case — the real `CASES` sweep to 256 MiB per buffer times four source chunks is a benchmark, not a test
- unknown-arm rejection, and every declared arm reachable from a branch in `main()`
- every `CASES` dtype round-tripping through the pool's allocator and torch's converter, on CPU — a case whose dtype the core cannot serve would otherwise only surface on a GPU box
- the jax arm on whatever backend is present, asserting the two device-transfer checks report **unknown** without a GPU rather than *safe*. A probe returning `ok` on a CPU box would retire the question it exists to keep open.
- table-driven tests for `Roundtrip.verdict` and `JaxCheck.poll_insufficient` — pure logic, no hardware requirement, no coverage before. Both encode a judgement worth pinning: that `is_pinned()` and the clock must agree before a round-trip counts as ok, and that only an async transfer JAX does *not* hold across condemns the pool's poll.

The GPU arms themselves (`h2d`, `overlap`, `roundtrip` timings) still need a GPU box and are not covered — the drift risk in them was the shared dtype mapping, which is now the core's and tested on CPU.

Full suite 266 passed / 12 skipped (was 241/11); ruff and mypy clean.

PR #32, opened 2026-08-16
Vectorize the per-sample random crop in the WB2 example

### LINKED ISSUE #29: Vectorize the per-sample random crop in `examples/wb2_dataloader.py`
### What would you like insitubatch to do?

`_subregion_crop` in `examples/wb2_dataloader.py:84-100` picks a random spatial crop per sample with a Python loop over the batch:

```python
out = np.empty(a.shape[:-2] + (h, w), dtype=a.dtype)
for b in range(a.shape[0]):
    i = int(rng.integers(0, lat - h + 1))
    j = int(rng.integers(0, lon - w + 1))
    out[b] = a[b, ..., i : i + h, j : j + w]
```

It should draw all the offsets at once and gather with array indexing, so the whole batch is one vectorized operation.

### Scope note, so this is not over-claimed

This is a **`batch_transform`**, not a `chunk_transform`. The project's stated invariant — a chunk transform must be vectorized numpy that releases the GIL — applies to the *decode* stage, where a Python loop serializes the pool. A batch transform runs after gather, and the docs name it as the home for per-sample random augmentation, which is inherently per-sample. **So this is not a violation of a documented rule**, and it should not be fixed as though it were.

What makes it worth doing is that it is an **example**, and examples get copied. `wb2_dataloader.py` is the file someone reads to learn how to attach a transform, and the pattern it currently teaches is a per-sample Python loop — the exact shape the project tells people to avoid one stage earlier. Someone who lifts it into a `chunk_transform` *has* introduced the bug.

### Does this touch a load-bearing invariant?

No — this fits within the current contract.

### Sketch, and the two parts that make it a real exercise

```python
i = rng.integers(0, lat - h + 1, size=B)     # (B,)
j = rng.integers(0, lon - w + 1, size=B)     # (B,)
rows = i[:, None] + np.arange(h)             # (B, h)
cols = j[:, None] + np.arange(w)             # (B, w)
```

1. **`a` is `(batch, ..., LAT, LON)` with arbitrary middle dims.** Advanced indexing with a leading batch axis, an ellipsis, and two trailing indexed axes needs care about broadcast shapes. `np.take_along_axis` applied twice, or `np.lib.stride_tricks.sliding_window_view` followed by a gather, are both reasonable — pick whichever reads most clearly, since this is example code whose job is to be read.

2. **The RNG draw order changes.** Today's loop draws `i, j` alternately per sample; a batched draw consumes the stream differently, so output will not be bit-identical for a given seed. That is fine — but a test must assert **shape, dtype and in-range-ness**, not equality with previously recorded values.

A comment noting that per-sample randomness is preserved (each sample still gets its own offset) is worth keeping, since that is the property a reader might otherwise think was lost.

### Related, out of scope

DESIGN.md notes that a `batch_transform` returning new arrays — `_subregion_crop` is the named example — drops its pooled buffer view immediately and gets no benefit from batch-buffer reuse. That is a separate design observation, not something this issue should try to fix.

### Not a candidate, already checked

`examples/wb2_arraylake.py:_crop` also contains a loop, but it iterates over `batch.arrays.items()` (2–3 variables) and applies **one** crop offset to the whole batch. Different semantics, not a hot-path concern — please leave it alone.

New to the project? Start with the [contributing guide](https://emfdavid.github.io/insitubatch/contributing/).

icn5381:
Closes #29.

## What changed

`_subregion_crop` in `examples/wb2_dataloader.py` no longer loops over the batch:

- the two offset vectors are drawn in one batched pull each (`rng.integers(..., size=B)`),
- middle dimensions fold into a flattened `(B*M, LAT, LON)` view, each sample's offsets repeat across its middle slices (so one spatial window per sample covers them all — same semantics as the loop),
- the crops are gathered with adjacent advanced indexing over every axis of the flattened view, then reshaped back to `(batch, ..., h, w)`.

I first tried broadcasting `(B, 1, h, 1)` / `(B, 1, 1, w)` index blocks against trailing `a[..., r, c]`, but NumPy appends the full broadcast block after `a.shape[:-2]` rather than merging the leading axes — shape came out wrong with middle dims present. The fold-and-gather form is the one I verified against a brute-force reference. Per-sample randomness is unchanged (each sample its own offset); only the RNG consumption order differs, so a given seed is not bit-identical to the old loop — as the issue anticipated, which is why the tests assert properties, not recorded values.

## How I verified

New `tests/test_wb2_subregion_crop.py` (5 tests): full-size crop is the identity; shape/dtype preserved through middle dims; every crop is a real contiguous window of its own sample (located by search, so a transposed or hoisted axis fails); constant-filled samples prove no cross-sample bleed; same seed reproduces, different seed differs. Plus the existing `test_example.py` smoke tests still pass, and the full suite is green: 241 passed, 29 skipped (`tests/test_zarr_indexing_parity.py` errors at collection on this machine — missing optional `zarr_indexing` module, pre-existing and unrelated). `ruff check` and `ruff format` clean on the touched files.

### DIFF
--- a/tests/test_wb2_subregion_crop.py
+++ b/tests/test_wb2_subregion_crop.py
@@ -1,0 +1,80 @@
+"""Shape / dtype / in-range-ness tests for the vectorized `_subregion_crop`.
+
+The batched RNG draw consumes the stream in a different order than the old
+per-sample loop, so output for a given seed is not bit-identical to the
+previous implementation — these tests assert the properties that must hold
+instead of recorded values (#29).
+"""
+
+from __future__ import annotations
+
+from types import SimpleNamespace
+
+import numpy as np
+
+from examples.wb2_dataloader import _subregion_crop
+
+
+def _apply(a: np.ndarray, subregion: tuple[int, int], seed: int) -> np.ndarray:
+    batch = SimpleNamespace(arrays={"t2m": a})
+    return _subregion_crop("t2m", subregion, seed)(batch).arrays["t2m"]
+
+
+def _find_window(sample: np.ndarray, crop: np.ndarray) -> tuple[int, int] | None:
+    """Locate the (i, j) whose window equals the crop, or None if none does."""
+    lat, lon = sample.shape[-2], sample.shape[-1]
+    h, w = crop.shape[-2], crop.shape[-1]
+    for i in range(lat - h + 1):
+        for j in range(lon - w + 1):
+            if np.array_equal(sample[..., i : i + h, j : j + w], crop):
+                return (i, j)
+    return None
+
+
+def test_full_size_crop_is_identity_3d():
+    a = np.random.default_rng(1).standard_normal((4, 6, 8))
+    out = _apply(a, (6, 8), seed=0)
+    assert out.shape == (4, 6, 8)
+    assert np.array_equal(out, a)
+
+
+def test_shape_and_dtype_preserved_with_middle_dims():
+    rng = np.random.default_rng(2)
+    a = rng.standard_normal((5, 3, 7, 9)).astype("float32")  # (B, T, LAT, LON)
+    out = _apply(a, (4, 5), seed=3)
+    assert out.shape == (5, 3, 4, 5)
+    assert out.dtype == a.dtype
+
+
+def test_each_crop_is_a_real_contiguous_window_of_its_own_sample():
+    # Distinct values per sample expose any cross-sample bleed: sample b is
+    # filled with the constant b, so a crop that mixed samples would show
+    # more than one value in a slice.
+    rng = np.random.default_rng(4)
+    a = rng.standard_normal((6, 2, 10, 12))
+    out = _apply(a, (5, 7), seed=5)
+    for b in range(a.shape[0]):
+        window = _find_window(a[b], out[b])
+        assert window is not None, f"sample {b}: crop is not a window of its own sample"
+
+
+def test_per_sample_offsets_are_independent():
+    # Constant-fill samples: every value in out[b] equals b, and at least two
+    # samples land on different windows when the domain is large enough for
+    # the seeds to differ.
+    a = np.zeros((8, 10, 10))
+    for b in range(a.shape[0]):
+        a[b] = b
+    out = _apply(a, (3, 3), seed=6)
+    for b in range(a.shape[0]):
+        assert np.all(out[b] == b), f"sample {b} contains values from another sample"
+
+
+def test_same_seed_is_reproducible_and_different_seed_differs():
+    rng = np.random.default_rng(7)
+    a = rng.standard_normal((4, 9, 9))
+    out_a1 = _apply(a, (4, 4), seed=42)
+    out_a2 = _apply(a, (4, 4), seed=42)
+    out_b = _apply(a, (4, 4), seed=43)
+    assert np.array_equal(out_a1, out_a2)
+    assert not np.array_equal(out_a1, out_b)
--- a/tests/test_wb2_subregion_crop.py
+++ b/tests/test_wb2_subregion_crop.py
@@ -59,15 +59,17 @@ def test_each_crop_is_a_real_contiguous_window_of_its_own_sample():
 
 
 def test_per_sample_offsets_are_independent():
-    # Constant-fill samples: every value in out[b] equals b, and at least two
-    # samples land on different windows when the domain is large enough for
-    # the seeds to differ.
-    a = np.zeros((8, 10, 10))
-    for b in range(a.shape[0]):
-        a[b] = b
+    # Random content makes every sample's window locatable (_find_window),
+    # so the offsets can be inspected, not just inferred. Requiring at
+    # least two distinct windows fails the degenerate form where a single
+    # scalar offset is broadcast across the whole batch — the exact bug a
+    # vectorized rewrite could introduce (all samples cropped identically).
+    rng = np.random.default_rng(8)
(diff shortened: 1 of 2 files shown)


tests/test_wb2_subregion_crop.py
```diff
@@ -0,0 +1,80 @@
...
+    # more than one value in a slice.
+    rng = np.random.default_rng(4)
+    a = rng.standard_normal((6, 2, 10, 12))
+    out = _apply(a, (5, 7), seed=5)
+    for b in range(a.shape[0]):
+        window = _find_window(a[b], out[b])
+        assert window is not None, f"sample {b}: crop is not a window of its own sample"
+
+
+def test_per_sample_offsets_are_independent():
```
emfdavid, line 61: This one can't fail on what its name promises, and it's worth fixing because the property is the whole point of the change.

I replaced your two batched draws with a single scalar offset broadcast across the batch — every sample cropped at the identical window, which is exactly the bug vectorizing this could introduce — and all five tests still passed. This one included: constant-filling sample b makes every window of it identical, so the offset is unobservable by construction. What survives is a second cross-sample-bleed check, which test_each_crop_is_a_real_contiguous_window_of_its_own_sample above already does more strictly.

The comment on lines 62-64 describes the test you meant to write ("at least two samples land on different windows"), and _find_window already returns the (i, j) it locates — it's just discarded at line 57. Something like: keep the located offsets from real (non-constant) data and assert the batch produced more than one distinct one. Fixed seed and real values, so it stays deterministic rather than probabilistic.

Small separate thing: the "sample b is filled with the constant b" comment on lines 50-52 belongs to this test, not the one it's sitting on.

emfdavid (commented):
Thank you for picking this up — and for writing up the broadcasting approach that *didn't* work. That saved me from suggesting it.

The tests are the strongest part. `_find_window` comparing the whole sample, middle dims included, pins "one window per sample covers all of its middle slices" — I broke that on purpose (swapping `np.repeat` for `np.tile`) and your test caught it. That's a better bar than the issue asked for. I've left one comment inline on a gap in the same file.

**The main thing: worth benchmarking against the sliding-window-view form.** SWV was new to me too, but the difference is bigger than I expected. Your gather indexes *every* axis with an array, so numpy walks it element by element and the window's contiguity buys nothing. Leaving the last two axes as slices lets it copy each window as a block:

```python
win = np.lib.stride_tricks.sliding_window_view(a, (h, w), axis=(-2, -1))
out = win[np.arange(a.shape[0]), ..., i, j, :, :]
```

The ellipsis separates the advanced indices from the trailing `:, :`, so the batch axis lands first and the middle dims stay put — the reshape goes away too. What I measured (4 vCPU GCP box, numpy 2.4.6 / py3.12.13, float32, best of 5x200, all three forms checked against a brute-force reference first — script: https://gist.github.com/emfdavid/6daa67d2d93360bef9031b648de7ed22):

```
case                                              loop    this PR       swv
WB2 default    (32,64,128)    crop 16x16        0.033ms    0.048ms    0.017ms
WB2 demo-like  (32,64,128)    crop 48x48        0.053ms    0.315ms    0.040ms
wide batch     (256,64,128)   crop 16x16        0.269ms    0.343ms    0.059ms
middle dims    (64,8,121,240) crop 64x64        1.210ms   10.119ms    1.090ms
```

The gap widens with crop area, which is the per-element signature.

Would you run it on your machine and post what you get, with the setup line? I'd rather
have a second measurement than treat mine as settled — and if it doesn't reproduce for you, I want to know that before this merges.

To be clear: the issue told you to pick whichever form read most clearly and said nothing
about speed. That's on me — I didn't expect the performance difference.

Last thing, and it's a real question rather than a hint: I'm still tuning the PR template. Did it get in the way, or just feel like overkill for a change this size? The one part I do need is the author-attestation box — that's the whole basis of the AI policy — but the rest I'm happy to cut if it's noise.

icn5381:
Ran your script — it reproduces hard here, and the setup could hardly differ more from yours (Apple Silicon vs your 4 vCPU GCP box):

```
numpy 2.5.2 | python 3.14.6 | arm64

case                                              loop    PR #32       swv
WB2 default    (32,64,128)    crop 16x16        0.025ms    0.032ms    0.009ms
WB2 demo-like  (32,64,128)    crop 48x48        0.024ms    0.185ms    0.015ms
synthetic      (16,32,64)     crop 16x16        0.008ms    0.015ms    0.007ms
wide batch     (256,64,128)   crop 16x16        0.121ms    0.177ms    0.027ms
middle dims    (64,8,121,240) crop 64x64        0.426ms    6.370ms    0.471ms
```

Same shape, same ordering, and the gap widening with crop area. `8a85d50` switches the implementation to the sliding-window-view form essentially as you wrote it — batch axis as the array index, trailing axes as slices, no reshape. The full suite is green locally (370 passed / 30 skipped; `ruff check src tests bench examples` clean; the four mypy errors I see in `bench/` and `examples/hubble/` predate this change and don't touch these files).

Also fixed the independence test per your inline comment: constant-filled samples made every window of a sample identical, so a scalar offset broadcast across the batch passed all five tests. It now uses random content so each sample's window is locatable via `_find_window`, and requires at least two distinct windows — the broadcast-across-batch degenerate fails it.

On the template: it didn't get in the way — Summary / For reviewers / Checklist is roughly what I'd write anyway, and for a change this size it took two minutes. Keep it. The attestation box is the part I'd fight for too; for what it's worth it genuinely holds here.

emfdavid (approved):
:shipit:

PR #37, opened 2026-08-31
fix: make the chunk pool's "safe to take away" predicate true, not approximately true

### LINKED ISSUE #33: ChunkPool.fail() marks a slot ready while tile tasks are still writing it
`fail()` (`pool.py:645-656`) sets `slot.ready = True` on a slot with `remaining > 0` so a waiter stops hanging. But sibling tile tasks for the same chunk are still running (`scheduler.py:429-459`) — and on the non-`nan` path `_one` calls `fail()` and **returns without scattering**, so `remaining` never reaches 0.

Consequences:

- `ready` does not mean "assembled and nobody is writing", which is the premise eviction rests on (`_make_room`, `pool.py:560`: `k not in self._pinned and s.ready`). A poisoned slot is evictable while writers are live.
- The poisoned slot survives the epoch boundary — `unpin_all` drops only `not slot.ready` (`pool.py:531`). Next epoch `pin_if_ready` rejects it (`:420`) but `try_admit` takes the resident branch (`:375`) and returns `True` **without refetching**, so `wait_ready` re-raises the stale error forever while refetched tiles drive `remaining` negative (`:606`). That is a permanent hang, not just a stale cache entry.

Shape of the fix: separate *may a consumer stop waiting* (signalling — `error`) from *may the pool take the buffer away* (safety — no writers, no readers). Count outstanding tile **tasks**, decremented in a `finally` in `_one` regardless of outcome; drop a failed slot once it quiesces. Local design note: `design-notes/pool-slot-lifecycle.md`.

One of four issues sharing a root cause: `ChunkPool` spreads "is this slot safe to take away" across five loosely-coupled fields (`remaining`, `ready`, `claimed`, `error`, `_pinned`).

### LINKED ISSUE #34: ChunkPool.unpin_all() clears pins globally, corrupting a concurrent iteration
`unpin_all` does `self._pinned.clear()` (`pool.py:530`) and `_iterate` calls it at the start of **every** pass (`source.py:384`). Pins are per-iteration working state, but the map is pool-global.

Two producers over one `ChunkPool` is a supported configuration — `BatchBuffers`' own docstring names it (`buffers.py:238-247`), and `tests/test_pinned.py:198` / `tests/test_buffers.py:313` test against it. `.train`/`.val` are `_SplitView`s over the same pool (`source.py:289-306`).

Failure sequence for `zip(ds.train, ds.val)`:

1. `next()` on A starts A's producer; it admits and pins chunks.
2. `next()` on B runs B's `_iterate` prologue → `source.py:384` → `unpin_all` wipes **A's** pins, and `pool.py:531-532` drops A's in-flight partials.
3. A's producer is between `wait_ready` (`source.py:433`) and `gather` (`source.py:435`). Its chunks are now unpinned + ready, i.e. legal eviction victims (`pool.py:560`).
4. B's `try_admit` evicts one → A's `gather` raises `KeyError` (`pool.py:744`), or reads a slot mid-scatter if B re-admits the same key — likely, since cross-split chunk sharing is advertised at `source.py:361-363`.

`unpin_all`'s docstring justifies itself with "the prior scheduler is fully closed (no race)" — true for sequential epochs, false for two live iterations.

Fix: scope pins to the owning iteration — `dict[key, dict[OwnerId, int]]`; `unpin_all(owner)` releases only that owner's counts and drops only that owner's partials. Kept pool-level rather than on the slot because `pin_keys` must record a pin for a not-yet-allocated key (`pool.py:484`). Local design note: `design-notes/pool-slot-lifecycle.md`.

Sibling of #33; same root cause.

### LINKED ISSUE #35: _Slot.claimed is a single bool, so one iteration's claim satisfies another's wait_ready
`claimed` is one bool per slot (`pool.py:182`), but it means "*the driver* referenced it this epoch" — and with two iterations over one pool there are two drivers.

`wait_ready` blocks on `ready and claimed` (`pool.py:700-705`). `pin_if_ready` sets `claimed = True` (`:426`), as does `try_admit` (`:377`, `:392`).

With `zip(ds.train, ds.val)` sharing a chunk:

1. B's `pin_if_ready` pins the chunk and sets `claimed = True`.
2. A's `wait_ready` sees `ready and claimed` — satisfied by **B's** claim — and returns, though A never referenced the slot.
3. A gathers it, then releases it at last use: `unpin_block` → `unpin_keys` (`source.py:448`) decrefs a key A never increfed, i.e. it decrements **B's** count.
4. B's in-use chunk drops to refcount 0 and becomes evictable mid-use.

This defeats the exact ordering `claimed` exists to guarantee (pin-before-consume-before-release, `pool.py:684-690`) — it is correct for one iteration and wrong for two.

Fix falls out of owner-scoped pins (#34): `claimed` becomes `refs[key][owner] > 0`, which is what the driver's pin already establishes and which is per-owner by construction. The flag can then be deleted. Local design note: `design-notes/pool-slot-lifecycle.md`.

Sibling of #33 and #34; same root cause. Found while tracing that plan, not by review.

### LINKED ISSUE #38: Auto-sized residency budget assumes one iteration, so concurrent iterations deadlock
`InSituDataset` auto-sizes `cache_budget_bytes` from one iteration's working set
(`source.py:216-260`), but one `ChunkPool` is deliberately shared by every active
iteration — `zip(ds.train, ds.val)` and two DataLoaders are supported and documented
(`pool.py:299-302`, `buffers.py:238-247`, `tests/test_pinned.py:198`). Two concurrent
iterations each need their working set resident at the same time, so the default is
too small for a configuration we advertise, and the failure is a hard stop:

```
RuntimeError: residency budget exhausted: cannot admit chunk 24 of 't2m'. The pool
holds 20 chunk(s) (2560 of 2560 bytes), every one of them pinned or in flight; no tile
is in flight; and the consumer is blocked waiting on [('t2m', 0)].
```

### Why this surfaced now

It was masked. `unpin_all()` cleared the pin map globally, so starting a second
iteration spuriously freed the first one's pins and made room — that is #34. With
references correctly owner-scoped (PR #37), both iterations legitimately hold their
pins and the real requirement shows up.

**This is not a leak.** Measured on a 32-chunk store (`spc=8`, `block_chunks=2`,
`batch_size=8`):

| budget | one iteration | two iterations |
|---|---|---|
| 12 chunks | OK (peak pinned 12) | deadlock |
| 20 chunks | OK (peak pinned 20) | deadlock |
| 32 chunks | OK | OK (peak pinned 32) |

A single iteration works at every budget and grows peak-pinned to whatever it is
given — read-ahead saturating the budget is by design. Only the concurrent case needs
more than it is handed.

### What to decide

1. **Should the default account for concurrent iterations at all?** It cannot know how
   many there will be, so any automatic answer is a guess. Options: size for one and
   document the requirement; size for two; or expose the per-iteration working-set
   figure so callers can multiply it themselves.
2. **The starvation error should name this cause.** It currently advises "raise
   cache_budget_bytes, or lower batch_size / block_chunks", which is right but does
   not mention that a second concurrent iteration doubles the requirement — the most
   likely trigger now that pins are scoped correctly.
3. **`docs/tuning.md` needs the guidance** once the sizing policy is settled. PR #37
   deliberately left that checklist item unticked rather than document a number that
   is about to change.

Adjacent but distinct from #33/#34/#35: those were correctness of the eviction
predicate; this is sizing policy on top of a now-correct predicate.

emfdavid:
## Summary

Closes #33, closes #34, closes #35 — three bugs with one root cause. Also closes #38
(the sizing decision and the diagnostic that follows from it; second commit). Eviction eligibility
was `k not in self._pinned and s.ready`, spread across five loosely-coupled fields
(`remaining`, `ready`, `claimed`, `error`, `_pinned`), and each issue is a different field
lying:

- **#33** — `fail()` had to set `ready = True` to wake a waiter, because that was the only
  lever available. That simultaneously declared a half-written slot a finished cache entry
  while sibling tile tasks were still writing into it, *and* left the poisoned slot resident,
  so the next epoch took `try_admit`'s resident branch and re-raised the stale error forever
  instead of refetching.
- **#34** — `unpin_all()` cleared the pin map globally. With two producers over one pool
  (`zip(ds.train, ds.val)`, documented in `buffers.py:238-247`), one iteration's epoch
  boundary stripped the other's pins and its in-use chunks became eviction candidates
  mid-gather.
- **#35** — `claimed` was a single bool, so one iteration's claim satisfied another's
  `wait_ready`. That iteration then gathered a chunk it never referenced, and its release
  decremented someone else's count.

All three produce **plausible** data, which throughput, shapes and smoke tests all pass —
only byte fingerprints catch them.

### The shape

A slot now carries one explicit `SlotState` (`FILLING → ASSEMBLED → READY`, `FAILED`
terminal) advanced in exactly one place (`_advance`), plus two counters that answer one
question each:

- `writers` — tile tasks **running**, so eviction is never racing a live write.
- `pending` — tiles not yet delivered, so completeness is separate from quiescence.

References become owner-scoped (`dict[key, dict[owner, int]]`) and **a reference *is* that
owner's claim**, so `claimed` is deleted. One iteration = one owner, minted by `_iterate` and
shared with its scheduler (`Scheduler.owner`).

Every tile write happens inside `pool.tile_write`, a scope that releases on **every** exit
path — including cancellation at an `await`, which an early `break` triggers via
`cancel_futures=True` and which no explicit call site can cover.

`unpin_all()` splits into `release_owner()` and `reset_epoch_counters()`. Releasing
references and zeroing counters are unrelated jobs on different clocks, and bundling them is
*why* the reset had to unpin globally. Cleanup also moves into `_iterate`'s own teardown, so
an abandoned pass cleans up after itself rather than relying on the next pass's prologue —
which, now that owners are scoped, would be a different owner and could not.

## For reviewers

**Behavior is meant to be unchanged for a single iteration.** The one intended behavioral
change is for concurrent iterations, below.

Most valuable second look, in order:

1. **`ChunkPool._advance`** — the only function that moves state. Worth checking the
   quiesced-but-incomplete case (a cancelled pass leaves an abandoned partial: stays
   `FILLING`, dropped by `release_owner`) and that a `FAILED` slot is deliberately *not*
   dropped there, because a consumer still has to observe the error.
2. **`ASSEMBLED` is load-bearing, not ceremony.** The chunk transform and the persist
   write-back run *outside* the lock between the last tile landing and publication, and
   another thread can observe that window. A predicate derived from a tile counter alone
   would call it evictable. This is the one place the design note's proposed
   `assembled = writers == 0 and error is None` would have introduced a bug.
3. **`writers` counts *started* tasks, not planned ones.** Initialising it to the tile count
   at admit is wrong: a cancelled pass leaves tiles never started, so the slot would never
   quiesce and its budget would never be reclaimable.
4. **`pin_keys` now notifies.** Now that a reference is the claim, pinning can be the event
   that satisfies a parked `wait_ready`; under the old shared flag it never could.

Two notes on the tests:

- The original #34 reproducer asserted that A's pins survive B's prologue. **It was not a
  valid discriminator** — A's own producer releases those pins in the background within
  ~0.3 s whether or not B ever starts (measured), so it would fail with the bug *fixed*. It
  is replaced by a deterministic pool-level test of owner-scoped release; the byte-fingerprint
  test remains the end-to-end statement.
- `deliver_tile` is sugar over `tile_write`, not a second lever — it opens the same scope, so
  the writer count still moves in exactly one place.

### Intended behavioral change: concurrent iterations need a larger budget

Two concurrent iterations now need a budget covering **both** working sets; the concurrent
test went from 20 → 32 chunks. This is not a leak — verified that a *single* iteration grows
peak-pinned to whatever budget it is given and works at every size, unchanged. The old global
unpin was spuriously freeing the other iteration's pins, so any budget that "worked" for
`zip(ds.train, ds.val)` before was relying on the bug.

The auto-sized default is still sized for one iteration, **and that is now the settled
decision** (#38): the engine cannot know how many iterations a caller intends to run, so an
automatic multiplier would silently cost memory in the single-iteration case that is almost
every case. Running several is an explicit choice, so sizing for it is the caller's.

The second commit makes that decision liveable, since the whole cost then falls on the
diagnostic. Starvation previously advised *"raise cache_budget_bytes, or lower batch_size /
block_chunks"* — correct, but not actionable when **every** resident chunk is legitimately
referenced and the caller cannot see why. It now names the cause:

```
... Nothing can free a slot, so this would hang. The working set is larger than the budget it
was sized for -- raise cache_budget_bytes, or lower batch_size / block_chunks. NOTE: 2
iterations are sharing this pool (e.g. `zip(ds.train, ds.val)`, or two DataLoaders); each
needs its own working set resident at once, so the budget must cover all 2.
```

Plus `docs/tuning.md` §"Several iterations at once multiply the budget" with the sizing rule
and why the default is not multiplied, and the same caveat at the auto-sizing site and on the
docstring that advertises the shared-pool configuration — so it is not carried by prose alone.
No example uses `zip(ds.train, ds.val)` as a runnable pattern; the mentions are the docstrings
and tests that advertise it, which is where a reader learns it is supported, so those are what
carry the caveat.

**`active_owners` counts from mint to release, not from first pin.** My first implementation
counted owners currently holding a reference and the note never fired — the second iteration
starves *before* it pins anything, which is exactly the case the message exists for. The test
caught it.

## Author attestation

**Left unchecked deliberately — this was drafted by Claude, and that box is the human
author's to tick.**

## Checklist

- [x] Tests added or updated — `tests/test_pool_lifecycle.py` reproduces all three; confirmed
      failing before the change for the documented reasons
      `uv run pytest -q` are green locally (313 passed, 12 skipped)
- [x] User-facing behavior documented in `docs/*.md` — **not done.** The concurrent-iteration
      budget requirement is user-facing and belongs in `docs/tuning.md`; deferred to #38 so the
      guidance and the auto-sizing fix land together rather than documenting a number that is
      about to change.
- [x] No load-bearing invariant is broken — no hot-path change (the state transition is the
      same lock sections the completion counter already used), `Batch` untouched, parallelism
      still in the event loop
- [x] Touches `ChunkPool`, the scheduler, and cross-thread readiness → free-threaded run
      passes: 80 passed, 1 skipped on 3.13t with `PYTHON_GIL=0` (pool, lifecycle, scheduler,
      residency, prefetch, window); re-run after the second commit, 70 passed / 1 skipped
- No performance claim.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,32 @@
 
 ## Unreleased
 
+- **The chunk pool's "safe to take away" predicate is now true, not approximately true.**
+  Eviction eligibility was spread across five loosely-coupled fields, and each one could
+  lie. `fail()` had to set `ready = True` to wake a waiter — the only lever available —
+  which simultaneously declared a half-written slot a finished cache entry while sibling
+  tile tasks were still writing into it, and left the poisoned slot resident so the *next*
+  epoch re-raised the stale error forever instead of refetching (#33). `unpin_all()`
+  cleared the pin map globally, so with two producers over one pool (`zip(ds.train,
+  ds.val)`, a documented configuration) one iteration's epoch boundary stripped the
+  other's pins and its in-use chunks became eviction candidates mid-gather (#34). And
+  `claimed` was a single bool, so one iteration's claim satisfied another's `wait_ready`
+  — that iteration then gathered a chunk it never referenced and its release decremented
+  someone else's count (#35). All three produced *plausible* data, which throughput,
+  shapes and smoke tests all pass; only byte fingerprints catch them.
+  A slot now carries one explicit `SlotState` (`FILLING → ASSEMBLED → READY`, `FAILED`
+  terminal) advanced in exactly one place, plus two counters that answer one question
+  each: `writers` (tile tasks *running*, so eviction is never racing a live write) and
+  `pending` (tiles not yet delivered, so completeness is separate from quiescence).
+  References are owner-scoped, and a reference *is* that owner's claim, so `claimed` is
+  gone. `Scheduler` takes the pool's obligation off the caller: every tile write happens
+  inside `pool.tile_write`, whose scope releases on **every** exit path — including
+  cancellation at an `await`, which no explicit call site can cover.
+  `ASSEMBLED` is a real state, not a formality: the chunk transform and the persist
+  write-back run outside the lock between the last tile landing and the slot being
+  published, and a predicate derived only from a tile counter would call that window
+  evictable.
+
 - **Added: `insitubatch.print_debug_info()` — one paste instead of a dozen version questions.**
   Reports the storage stack (zarr / obstore / numpy / xarray), whichever framework adapter is
   actually installed, and the **free-threading state** — both the build flag and whether an
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 ## Unreleased
 
+- **Under-sizing the budget for concurrent iterations now says so.** One `InSituDataset`
+  owns one chunk pool and every active iteration shares it — `zip(ds.train, ds.val)`, or
+  two `DataLoader`s — but each holds its *own* chunk references, so residency is the sum
+  of their working sets, not the maximum. The auto-sized default covers one iteration and
+  deliberately stays that way: the engine cannot know how many you intend to run, and
+  guessing high would cost memory in the single-iteration case that is almost every case.
+  What was missing was the diagnostic. Starvation previously advised "raise
+  cache_budget_bytes, or lower batch_size / block_chunks" — correct but not actionable
+  when *every* resident chunk is legitimately referenced and the caller has no way to see
+  why. It now names how many iterations are sharing the pool, and the pattern that
+  produces that. Owners count from mint to release rather than from their first pin,
+  because the iteration that starves before it can pin anything is exactly the one that
+  needs naming. `docs/tuning.md` gains the sizing rule.
+
 - **The chunk pool's "safe to take away" predicate is now true, not approximately true.**
   Eviction eligibility was spread across five loosely-coupled fields, and each one could
   lie. `fail()` had to set `ready = True` to wake a waiter — the only lever available —
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -171,10 +171,13 @@ As built (PR #4), the engine stayed simpler than the original sketch:
   give disjoint train/val labels for any offset (injective translation), and the shared
   boundary timestep is only ever a val *input*, never scored. Task meaningfulness
   (non-degenerate offset choices) is the **user's** responsibility, not the engine's.
-- Residency is **reference-counted** (`pool._pinned: dict[key,int]`, not a boolean set)
-  because a windowed chunk can feed several blocks; a per-epoch `claimed` flag orders
-  pin→consume→release across the epoch boundary (fixes a cross-epoch gather/leak race).
-  Decode-once across views holds (slots key on `path`).
+- Residency is **reference-counted and owner-scoped** (`pool._pinned: dict[key, dict[owner,
+  int]]`, not a boolean set) because a windowed chunk can feed several blocks *and* because
+  one pool serves several concurrent iterations. A reference **is** that owner's claim, which
+  orders pin→consume→release across the epoch boundary (fixes a cross-epoch gather/leak race)
+  and, being per owner, cannot be satisfied by another iteration's claim. One iteration = one
+  owner, minted by `_iterate` and shared with its scheduler. Decode-once across views holds
+  (slots key on `path`).
 - **Splits are iterable views, not a constructor arg.** `InSituDataset` is split-agnostic;
   you iterate `ds.train` / `ds.val` / `ds.test` / `ds.all`, all sharing **one** `ChunkPool`.
   This matters *because* of windows: a windowed read near a split boundary spills into the
@@ -665,7 +668,7 @@ Things wrong or missing in *our* code today, with the reasoning that sets their 
 - **Refcount overhead on the non-windowed path (~3.5%, accepted).** Refcounted residency +
   offset-aware `gather` cost ~3.5% at the **GRIB-end local worst case** (one sample/chunk,
   ~36 KB blobs where decode is ≈ free, GIL build) — not residency (`resident` unchanged) and
-  not per-block setup; it is per-batch gather + per-chunk `dict`/`claimed` bookkeeping under
+  not per-block setup; it is per-batch gather + per-chunk reference bookkeeping under
   `pool._cv`. It shrinks toward noise at real chunk sizes and is invisible on blob stores.
   The microscope for this and any future *locking* work is a tiny local store with negligible
   decode under free-threading (`PYTHON_GIL=0`, where `pool._cv` is the only contention
--- a/docs/tuning.md
+++ b/docs/tuning.md
@@ -95,6 +95,37 @@ If you set `cache_budget_bytes` above the working set, residency rises to that b
 purpose — that extra memory *is* the cross-epoch cache. Point `cache_dir` at local NVMe to
 spill it to disk instead of RAM.
 
+### Several iterations at once multiply the budget
+
+The three bounds above describe **one** iteration. One `InSituDataset` owns one chunk pool,
+and every active iteration shares it — `zip(ds.train, ds.val)`, or two `DataLoader`s over the
+same dataset. That is supported, and chunks a windowed read pulls across a split boundary are
+decoded once and reused by both. But each iteration holds its **own** references to the chunks
+it is working on, so residency is the sum, not the maximum:
+
+```
+cache_budget_bytes  >=  n_concurrent_iterations x (block_chunks x outer_chunk_bytes)
+```
+
+**The auto-sized default is deliberately computed for one iteration, and stays that way.**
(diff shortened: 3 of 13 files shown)


emfdavid (commented):
Looks clean

Ran:
`PYTHON_GIL=0 uv run --python 3.13t python -m bench.probe_decode   --max-chunks 64 --repeats 5 --decode-threads 1,2,4,8 --no-raw` and `uv run python -m bench --engines insitu,memory --chunk-sizes 1 --epochs 2 --repeats 5` - no regression

emfdavid:
Pushed a second commit (\`0ce254e\`) after your review — flagging since it lands on an already-reviewed PR.

**Decision recorded:** size for one is the correct default, so #38 needed no auto-sizing change. That puts the whole cost on the diagnostic, which is what this commit fixes: the starvation error now names how many iterations share the pool and the pattern that produces it, `docs/tuning.md` gains the sizing rule and the reason the default is not multiplied, and the caveat also sits at the auto-sizing site and on the `BatchBuffers` docstring that advertises the shared-pool configuration.

No example uses `zip(ds.train, ds.val)` as a runnable pattern — `examples/advection/train_tf.py:80` is `zip(grads, ...)` for TF gradients, unrelated. The mentions are docstrings and tests that advertise it as supported, which is where a reader learns it, so those carry the note.

One implementation note: `active_owners` counts from **mint to release**, not from first pin. My first version counted current pin-holders and the note never fired — the second iteration starves *before* it pins anything, which is exactly the case the message exists for. The test caught it.

Verified: 299 passed / 20 skipped against a 298 / 20 baseline on the same tree and extras (delta is exactly the new test); ruff, mypy, `mkdocs --strict` clean; free-threaded 70 passed / 1 skipped on 3.13t with `PYTHON_GIL=0`. This closes #38, so the PR now closes #33, #34, #35, #38.

PR #49, opened 2026-09-03
docs: correct claims stronger than the mechanism, and stale ones the code outgrew

### LINKED ISSUE #47: Docs: cut imprecise claims, then simplify
Some claims in the docs are stronger than the mechanism supports, or have been outgrown by the code; three concrete examples are in a comment below. Fix those, then sweep for the same two failure modes, preferring the narrower true statement over the stronger one, because the narrow one survives a reader who knows the field. Separately, the pages have accumulated overlap and length that a reader has to work through before reaching the argument, so this is a pass to cut and reorganize as much as to correct.

**Style worth copying:** [LanceDB's data-loading guide](https://www.lancedb.com/blog/data-loading-guide) (Weston Pace, Aug 2026). Diagrams captioned with the mechanism rather than a restatement of the picture; before/after pairs that carry the argument; numbers embedded in the figures as units; a performance target derived from napkin math instead of asserted.

Related: #40.

emfdavid:
## Summary

Docs only. Fixes the three defects in #47, then sweeps the front doors for the same two failure modes: a claim stronger than the mechanism supports, and a claim the code has outgrown.

Closes #47.

Found while drafting the Pangeo Showcase abstract, which is why these are the ones an outside reader meets first.

**The three from the issue.** "Sync `getitem` can't drive async obstore" is false — zarr-python's sync API runs a process-wide event loop, so `arr[...]` inside a worker drives async obstore fine. The true claim is scope, not capability: `__getitem__` returns one sample before the next starts, so fan-out never spans the samples a batch needs, and the budget is per-process. That claim had spread to **six sites across four files**. "Batch buffers are not pooled … host memory is not pinned yet" has been stale since #21 and offered the shipped work as "one deferred fix". "No copy — `gather` views the slot in place" is false at three sites; `gather` allocates from the buffer pool and copies each chunk's rows in.

**The sweep.** "~ms to first batch" (README, landing page) was off by ~10× — measured TTFB is 0.2 s at c1, 0.7 s at c16, ~320 ms on WB2. The README Status block cited a run that does not match the benchmarks page. "Free-threading-correct by construction" was one claim covering two mechanisms with different guarantees. `read-once` claimed cross-epoch reuse the default budget does not give.

## For reviewers

**The README Status block is what most needs a second look.** It claimed `721×1440` fields where `docs/benchmarks.md` says coarsened `361×720`, and "~8× the throughput" where the page reports ~8× **memory** and ~19× **throughput** — which looks like a transposition. I realigned every figure to the published c16 table (`docs/benchmarks.md:230-244`), so the throughput claim goes **up**, 8× → 19×, and added the caveat that this is the fat end of the sweep where a tuned pool loses hardest. **If those original numbers came from a real run that simply is not on the benchmarks page, this replaces real data with different real data.** You would know that; I would not.

Second: the free-threading split in the README. "The engine is free-threading-correct by construction" now applies only to the `ChunkPool` scatter, which earns it structurally. The batch-buffer pool is described separately as *enforced* — it reads `sys.getrefcount` against a calibrated baseline, which off-GIL can be stale, and a below-baseline count raises rather than lending. That distinction is not documented anywhere in the repo today, and I wrote only what `buffers.py` and `DESIGN.md:565-575` support. No measured cost of stale-high readings is claimed, because there is no in-repo measurement to cite.

Confident in the rest: the six-site defect-1 rewrite, the buffer-pool paragraph (verified against `buffers.py:261` and `frameworks.py:281`), and the `gather` copy claims (verified against `pool.py:1226`, `out[(mask, *dst[1:])] = tile[(rows, *src[1:])]`).

**Rebased onto #41** after it landed mid-review. Two conflicts in `README.md`, both resolved by keeping #41's newer mechanism wording and re-applying this branch's correction on top: the pool sentence now reads "holds them and assembles batches on demand" (#41's), and the free-threading paragraph keeps #41's "each tile is its own key, so writers never collide" while taking this branch's split between the `ChunkPool` claim and the batch-buffer pool's. Every edit was re-verified against post-#41 code.

## Author attestation

## Checklist

- [x] `uv run ruff check src tests bench examples`, `uv run mypy src bench examples` are green locally; `uv run --extra docs mkdocs build --strict` is clean
- [ ] A bullet added under `## Unreleased` in `CHANGELOG.md` — **deliberately skipped** at the maintainer's direction; this corrects existing docs rather than changing behavior
- [x] No load-bearing invariant is broken — docs only, no code touched

---

https://claude.ai/code/session_011nhMRqZosrp81VjwkJn1zC

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -70,8 +70,10 @@ Three frictions with cloud Zarr:
 
 1. **No shared chunk cache** across workers → one chunk fetched + decompressed
    once *per worker* whose samples land in it.
-2. **Sync `getitem` can't drive async obstore** → you can't fan out 200 concurrent
-   range reads from inside a worker.
+2. **Fan-out is one sample deep** → a worker *can* drive async obstore (zarr's sync
+   API runs a process-wide event loop), but `__getitem__` returns one sample before
+   the next starts, so concurrency never spans the samples a batch needs — and the
+   budget is per-process, multiplied N times against the store.
 3. **dask thread pool nested in each worker** → procs × threads oversubscription,
    slow fork startup, fat memory.
 
--- a/README.md
+++ b/README.md
@@ -35,9 +35,9 @@ spectra without building anything:
 
 The classic PyTorch `DataLoader` spreads work across worker **processes**, each
 running a *synchronous* `__getitem__`. Against cloud Zarr that means no shared
-chunk cache (every worker re-reads the same chunk), no way to drive async
-obstore, and dask thread pools nested inside forked workers. `insitubatch`
-**inverts** it: one async event loop streams stored chunks under a single
+chunk cache (every worker re-reads the same chunk), read concurrency that reaches no
+further than one sample, and dask thread pools nested inside forked workers.
+`insitubatch` **inverts** it: one async event loop streams stored chunks under a single
 concurrency budget into a bounded pool that holds them and assembles batches on
 demand — the pool doubles as the cache; torch runs `num_workers=0`.
 
--- a/README.md
+++ b/README.md
@@ -43,11 +43,11 @@ demand — the pool doubles as the cache; torch runs `num_workers=0`.
 
 The payoff is a **two-regime** story against the worker-process `DataLoader`. On a
 **well-chunked** store it **matches a hand-tuned worker/xbatcher pool** (swept to 32 workers)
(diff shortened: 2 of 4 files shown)


emfdavid (commented):
:shipit:

PR #54, opened 2026-09-04
feat: applies() -- per-variable transform scope, and cache identity per array

### LINKED ISSUE #42: cache_dir: no cross-process arbitration, and cache identity is not per-variable

emfdavid:
## Summary

Part 2 of #42, on top of #51 (writer arbitration) and #52 (the live `_persist` guard).
Follows the plan agreed in [the scoping comment](https://github.com/emfdavid/insitubatch/issues/42#issuecomment-5534685755);
no decisions were re-opened.

**This is a correctness fix before it is a caching one.** Gating a `chunk_transform` with
`if chunk.read.array == ...` inside its body is invisible to `output_geometry`, which folds
every transform's declared `output_inner` into **every** variable's geometry. A regrid gated
to `t2m` therefore made the engine believe `u10` was regridded too: `u10` was gathered as a
truncated prefix of itself, and its cache entry could never revive, because the `.npy` on
disk no longer matched the declared geometry. Nothing raised — right shape, right dtype,
wrong numbers.

Nothing in this tree was corrupt: all four gating sites (`kelvin_to_celsius`, Hubble's `SCI`,
SDSS's `flux`, `StandardScaler`'s implicit `KeyError`) are shape-preserving, and the one
shipped reshaping transform (`Coarsen`) is ungated. The hazard was in what a user writes next.

**Scope is now declared at the call site**, where the engine can read it:

```python
InSituDataset(..., chunk_transforms=[
    applies(["2m_temperature"], kelvin_to_celsius),   # this array only
    Coarsen(2),                                       # every array, as before
])
```

"Subtract 273.15" is a fact about Kelvin; "`2m_temperature` is in Kelvin" is a fact about
your store. The old pattern welded them, which is why one unit conversion had to enumerate
two archives' naming conventions (`TEMPERATURE_VARS`) and silently skipped a third. A
transform may be *parameterized* by variable — a fitted scaler's statistics legitimately are
— but it must not decide whether it runs.

`transforms_for(path, ...)` is the single resolution of scope; geometry, the fingerprint and
`_apply_transforms` all read it, so they cannot drift. Off-scope arrays skip the call
entirely. Unknown names raise at construction (before any cache dir or lockfile exists), and
a scoped `StandardScaler` is validated against its own stats dict there rather than
`KeyError`-ing in a decode thread.

**Manifest format 4: cache identity per array.** Each entry carries the hash of the
transforms scoped to *its own* array — on the entry, not in the header, because the log is
append-only and a header map cannot absorb an array a later run reads for the first time.
Editing `t2m`'s transform no longer invalidates `u10`/`v10`, whose bytes did not change.

Staleness becomes a partial answer, which is the part worth reviewing for semantics:

- the error **names which arrays**; `reset_stale_cache=True` deletes only their files;
- an array with **no entries** is *cold*, not stale — adding a variable is not a wipe;
- an array a run **does not read** keeps its entries and files, so two configurations can
  share one `cache_dir`;
- scope is deliberately **not** folded into a transform's own token, so narrowing a scope
  invalidates the variable that *left* and not the one that stayed;
- the log is compacted (temp file + `rename`, `_TMP_SUFFIX` so a crash is swept) when a load
  drops entries — safe only because #51's exclusive lock proves there is no second writer.

Only the log **format** is still all-or-nothing: a log we cannot parse says nothing about
any individual array.

Also: `ChunkTransform` had three definitions (a Protocol in `transforms.py`, a `Callable`
alias in `pool.py`, another in `check_transform.py`). There is now one — the Protocol, with
its `__call__` made positional-only so a plain function satisfies it whatever it names its
parameter.

Closes #42.

## For reviewers

**The semantics, not the code.** "Which arrays are stale" used to be all-or-nothing and is
now a subset; the error message is what makes that usable, and the three not-stale cases
above are judgement calls worth a second opinion — particularly "an array this run does not
read is left entirely alone", which is what lets two configurations share a directory but
also means we keep entries we cannot validate.

Confident in: the geometry fix (regression test drives the real `ChunkPool` and gathers), the
per-array fingerprint, and that a bare transform behaves exactly as before.

Deferred, as agreed in the issue: **per-path assembly** (a variable no transform touches
could stay on the tiled fast path, but `assembles` is a single bool the scheduler reads to
decide where delivery runs). Recorded in DESIGN.md's known limitations.

## Author attestation

*(Left unchecked deliberately — this PR was drafted with Claude Code, and the attestation is
the human author's to tick.)*

## Checklist

- [x] Tests added or updated — `tests/test_transform_scope.py`, 17 tests written first: the
      truncation repro, "editing `a`'s transform keeps `b`/`c`", cold-not-stale, an untouched
      array keeps its files, precise GC, scope narrowing, compaction (and a crashed one),
      `readonly_cache` rejection, unknown/empty/bare-string scopes, and the scaler check
      `uv run pytest -q` are green locally (435 passed, 6 skipped; TF run separately)
- [x] Docstrings and API docs for any new or changed public surface (`applies` is in
      `__all__`, so it lands in the generated API reference)
- [x] User-facing behavior documented in `docs/*.md` — a new "Scoping a transform to
      variables" section in `docs/architecture.md`, "What a transform edit invalidates" in
      `docs/tuning.md`, and the old per-pipeline limitation retired in both places + DESIGN.md
- [x] A bullet added under `## Unreleased` in `CHANGELOG.md` (two)
- [x] No load-bearing invariant is broken — transforms stay vectorized numpy, the hot path
      stays O(chunks) (scope is resolved once per path at construction, never per chunk)
- [x] Touches `ChunkPool` → the free-threaded (3.13t) run passes (confirmed in CI)
- [x] No performance claim made

https://claude.ai/code/session_01KnQ6EEXRwRSPKMAzUdgxpn

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -588,7 +588,8 @@ read → decode ─►[chunk_transform]─► buffer → gather ─►[batch_tra
    thread pool, **before** shuffle/gather. Amortized over every sample that draws
    from the chunk. Home for per-element, sample-order-independent ops: **scaling /
    normalization, unit conversion, dtype cast, chunk-local regrid.** Sees one
-   variable, one chunk (`chunk.read.array` gives the variable).
+   variable, one chunk. Restrict it to particular variables with `applies` — see
+   [Scoping a transform to variables](#scoping-a-transform-to-variables).
 2. **`batch_transform(Batch) -> Batch`** — per-batch, after gather. For ops that
    need the assembled batch: **cross-variable derived fields, channel stacking,
    per-sample random augmentation/crops, collation to model layout.**
@@ -609,6 +610,43 @@ Runnable side-by-side example:
 [`examples/transforms.py`](https://github.com/emfdavid/insitubatch/blob/main/examples/transforms.py)
 — a Kelvin→Celsius `chunk_transform` (one variable, cached) and a cross-variable windspeed
 `batch_transform` (needs the assembled batch, uncached).
+
+### Scoping a transform to variables
+
+`chunk_transforms` is one ordered list and a bare transform runs on **every** variable. To
+restrict one, wrap it at the call site with `applies`:
+
+```python
+from insitubatch import applies
+
+InSituDataset(..., chunk_transforms=[
+    applies(["2m_temperature"], kelvin_to_celsius),   # this variable only
+    Coarsen(2),                                       # every variable
+])
+```
+
+The names are zarr array **paths** (what `chunk.read.array` carries), not dict labels —
+two labels may alias one array (`t2m_now` / `t2m_next`), and the cache is keyed by path. A
+name matching no array raises at construction.
+
+**Do not gate inside the transform.** An `if chunk.read.array == "t2m"` in the body is
+invisible to the engine, and that is not a style point — it is a correctness bug for any
+reshaping transform. `output_geometry` folds every transform's declared `output_inner` into
+**every** variable's geometry; it cannot see a name test in a function body. A transform
+that halves `t2m` and no-ops on `u10` therefore makes the engine believe `u10` is halved
+too, and `u10` is gathered as a truncated prefix of itself, its cache file permanently
+unable to revive — with nothing raised. Right shape, right dtype, wrong numbers.
+
+The division of labour: "subtract 273.15" is a fact about Kelvin, "`2m_temperature` is in
+Kelvin" is a fact about your store. A transform may be *parameterized* by variable — a
+fitted `StandardScaler`'s statistics legitimately are, and it validates a declared scope
+against its own stats dict — but it must not decide whether it runs. Configuration inside,
+control flow outside. sklearn's `ColumnTransformer`, NVTabular's `cols >> Op()` and
+xarray's subset-then-`map` all place the selector the same way.
+
+Scope also buys two smaller things: the engine skips the call for arrays outside it (the
+no-op passes are gone), and cache invalidation becomes per array — see
+[the cache fingerprint](#cross-run-persistence).
 
 ### Standard scaler — pre-fit GLOBAL stats (not per-chunk)
 
@@ -875,11 +913,13 @@ ds = InSituDataset(store, manifest, cache_dir="/mnt/nvme/era5/v3",
 
 Two guards keep a reopened cache honest:
 
-1. a **chunk-transform fingerprint** in the log header — change your `chunk_transforms`
-   (or bump the log format) and the cache is **stale**. A stale cache is almost never
-   what you intended, so by default it **raises** at construction rather than silently
+1. a **chunk-transform fingerprint** stamped on every log entry, over the transforms
+   scoped to *that array* — change your `chunk_transforms` (or bump the log format) and
+   the arrays they affect are **stale**. A stale cache is almost never what you intended,
+   so by default it **raises** at construction, naming which arrays, rather than silently
    rebuilding or serving stale data. Pass `reset_stale_cache=True` to opt into deleting
-   the old files and rebuilding cold (or delete the `cache_dir` yourself); and
+   just those arrays' files and rebuilding them cold (or delete the `cache_dir`
+   yourself); and
 2. a per-entry **shape/dtype check** on revive — a chunk whose stored geometry no
    longer matches the current array is a **miss** (re-fetch + overwrite), never an error.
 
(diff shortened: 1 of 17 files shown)


src/insitubatch/pool.py
```diff
@@ -1474,11 +1545,15 @@ def set_error(self, error: BaseException) -> None:
             self._cv.notify_all()
 
     def _apply_transforms(self, array: str, chunk_index: int, data: np.ndarray) -> np.ndarray:
-        if not self._chunk_transforms:
+        """Run the transforms scoped to this array, in order. Off-scope transforms are not
+        called at all -- the no-op pass over a variable a transform does not understand is
+        gone, and so is any chance of it acting on one it was not given."""
+        pipeline = self._transforms_by_path[array]
+        if not pipeline:
```
emfdavid: Remove the description of the old behavior

docs/architecture.md
```diff
@@ -610,6 +611,43 @@ Runnable side-by-side example:
...
+
+InSituDataset(..., chunk_transforms=[
+    applies(["2m_temperature"], kelvin_to_celsius),   # this variable only
+    Coarsen(2),                                       # every variable
+])
+```
+
+The names are zarr array **paths** (what `chunk.read.array` carries), not dict labels —
+two labels may alias one array (`t2m_now` / `t2m_next`), and the cache is keyed by path. A
+name matching no array raises at construction.
```
emfdavid, line 630: This means you structurally can not apply different transform pipelines to labels backed by the same array. This needs to be stated clearly. I can't think of a real use case it violates that can't be solved with a batch transform anyway... but it should be a clear limitation in the docs.

src/insitubatch/pool.py
```diff
@@ -1474,11 +1545,15 @@ def set_error(self, error: BaseException) -> None:
             self._cv.notify_all()
 
     def _apply_transforms(self, array: str, chunk_index: int, data: np.ndarray) -> np.ndarray:
-        if not self._chunk_transforms:
+        """Run the transforms scoped to this array, in order. Off-scope transforms are not
+        called at all -- the no-op pass over a variable a transform does not understand is
+        gone, and so is any chance of it acting on one it was not given."""
```
emfdavid: Remove the description of the old behavior

tests/test_transform_scope.py
```diff
@@ -0,0 +1,442 @@
...
+
+    back = _persist_pool(_geoms(url), backing, [double])
+    owner = back.new_owner()
+    assert back.pin_if_ready("b", 0, owner), "and its manifest entry must survive too"
+    back.close()
+
+
+def test_narrowing_a_scope_invalidates_only_the_variable_that_left(three_vars, tmp_path):
+    """Scope is not folded into a transform's own token: the array that *stays* in scope
+    hashes identically, and only the one that left is stale."""
```
emfdavid: Is the one that left marked stale? That means it would be deleted and compacted?

emfdavid (commented):
Add/update unit tests for pipeline hash invariant behavior. List pointers if the coverage is already present.

Issue #57, opened 2026-09-05
Partial shard reads: fetch the inner chunks a run needs, not the whole shard


emfdavid:
Follow-up to #56, which makes sharded arrays *readable* by treating the shard as the stored
chunk. That is correct but coarse: we fetch and decode a **whole shard** whatever the run
actually needs.

## The amplification

zarr's sharding codec exists precisely so a reader does not have to do that — the shard index
gives the byte range of each inner chunk, so a reader can fetch only the inner chunks it wants.
We read whole objects by design (`scheduler.py:675`, `store.get(key)`), which on a sharded
array is the entire shard.

NOAA GFS analysis, the shape this shows up on:

| unit | shape | decoded size |
|---|---|---|
| inner chunk | `(1440, 50, 50)` | 14.4 MB |
| **shard** (what we fetch) | `(1440, 400, 400)` | **921 MB** |
| assembled outer chunk | `(1440, 721, 1440)` | 5.98 GB |

Measured with #56's fix: 8.88 GB peak RSS at `resident_peak=1`, 1440 samples in 6.9 s
(~870 MB/s). Throughput is fine; the cost is residency and the granularity of a miss.

**When it does not matter:** a run that consumes a whole outer chunk — which is the
train-in-place case, and the reason deep time-chunking is a *good* fit for us. 1440
consecutive samples out of a 1440-sample shard wastes nothing.

**When it does:** any run that wants a subset of a shard's spatial extent or a handful of its
samples. A user reading two timesteps pays 921 MB per shard to use ~0.14% of it. Region
subsetting (a CONUS window out of HRRR, a basin out of IMERG) is the obvious case, and it is
the normal way people use these archives.

## What changing it would mean

The stored-chunk unit stops being "one `store.get`". The read plan would work on the inner-chunk
grid, and a fetch becomes: read the shard index once, then byte-range reads for the inner chunks
this run needs. That touches the planner, the fetch path and the residency accounting, and it
interacts with:

- **#40** (naming: `tile` -> `stored_chunk`) — the vocabulary gets a third level (shard /
  inner chunk / our assembled chunk) and should be settled before, not after
- the chunked-pool slot direction — same question of what the buffer unit is
- caching: the persisted `.npy` is tile-major today, so the tile definition changing is a
  manifest-format question

Open design questions, not decided here:

1. Do we plan on inner chunks always, or only when the shard exceeds some size? A per-store
   heuristic is a second way to do the same thing, which we generally reject.
[...]

PR #58, opened 2026-09-06
fix: read the storage chunk, not the inner chunk, on sharded arrays

### LINKED ISSUE #56: Sharded zarr-v3 arrays fail to decode: we plan reads on the inner chunk, not the shard
Reading any **sharded** zarr-v3 array fails during decode with a checksum error that says
nothing about sharding:

```
zarr/core/chunk_utils.py:321 in decode_chunk -> zarr/codecs/crc32c_.py:47
ValueError: Stored and computed checksum do not match.
    Stored: b'\xbfW\x99\x94'. Computed: b'\x10L/\xef'.
```

The failing call is `index_transform.decode_chunk(index_bytes, index_spec)` — it is the
**shard index** that fails to verify, not the data. Plain zarr reads the same array fine, so
this is ours.

## Why it matters now

**Every dataset in the [dynamical.org](https://dynamical.org/catalog/) catalog is sharded** —
NOAA GFS/GEFS/HRRR/MRMS, ECMWF AIFS/IFS, DWD ICON-EU, NASA IMERG, ECCC HRDPS. That is one of
the larger public collections of analysis-ready weather zarr, and none of it is readable today.

WeatherBench2 and ARCO are **not** sharded, which is why every benchmark, example and test we
have ever run missed this.

We also already claim to support it. `docs/architecture.md:235-239` contrasts our chunk with
"a zarr **shard**" and defines the stored-chunk unit as "whatever a single `store.get`
returns — or one zarr shard on a sharded array". Meanwhile `grep -ri shard src/insitubatch/`
returns exactly one hit, and it is the word "reshard" in a docstring.

## Root cause

On a sharded array zarr reports two different shapes and only one of them is the storage unit:

| | NOAA GFS `temperature_2m` | meaning |
|---|---|---|
| `metadata.chunks` | `(1440, 50, 50)` | *inner* chunk — read granularity **inside** a shard |
| `metadata.chunk_grid.chunk_shape` | `(1440, 400, 400)` | the **shard** — what a chunk key addresses, what `store.get` returns |

We use `metadata.chunks` in three places, so we ask the store for a key believing it holds a
50x50 tile, get a 400x400 shard back, and hand that shard to a `ChunkTransform` whose head is
the `ShardingCodec` with an `ArraySpec` shaped `(1440, 50, 50)`. The codec sizes its index
from that spec, expects a 1x1x1 index, reads the wrong trailing bytes, and fails CRC.

The three sites:

- `src/insitubatch/store.py:154` — `ArrayGeometry(chunks=tuple(arr.chunks))`, which drives
  both `sample_chunk_size` and the inner tile grid
- `src/insitubatch/scheduler.py:576` — `ArraySpec(shape=meta.chunks, ...)`
- `src/insitubatch/scheduler.py:588` — `chunk_shape=tuple(aa.metadata.chunks)`

## Reproducer

Needs `insitubatch[icechunk]`; the bucket is anonymous, no credentials.

```python
import icechunk, insitubatch
st = icechunk.s3_storage(bucket="dynamical-noaa-gfs",
                         prefix="noaa-gfs-analysis/v0.1.0.icechunk",
                         region="us-west-2", anonymous=True)
store = icechunk.Repository.open(st).readonly_session("main").store
geoms = insitubatch.open_geometries(store, variables=["temperature_2m"])
m = insitubatch.split_by_chunk(geoms["temperature_2m"], fractions=(1.0, 0.0, 0.0),
                               sample_range=(1440, 1448))
ds = insitubatch.InSituDataset(store, m, geometries=geoms, batch_size=2,
                               block_chunks=1, shuffle=False)
next(iter(ds.all))          # ValueError: Stored and computed checksum do not match
```

A local sharded array reproduces it without network and is what the fix should be tested
against.

## Proposed fix

One spelling of "the shape of one stored object", used at all three sites:

```python
def storage_chunks(arr) -> tuple[int, ...]:
    return tuple(arr.metadata.chunk_grid.chunk_shape)
```

Identical on unsharded arrays, and **defined for zarr-v2 metadata too** (ARCO reports
`chunks == chunk_grid.chunk_shape == (1, 721, 1440)`), so the format-agnostic path is
unchanged.

Spiked on branch `spike/sharded-storage-unit` (commit `980d234`), ~26 lines including the
docstring. With it, batches from the reproducer above `np.allclose` against zarr's own read
of the same slice, and the suite is **446 passed, 6 skipped — unchanged**.

## What is still open before it ships

1. **Tests.** No sharded fixture exists. zarr writes one locally, so this needs no network:
   geometry reports the shard grid; a decode round-trip; the inner-tiled shape dynamical
   uses; and v2 unchanged.
2. **The byte budget with fat tiles.** One GFS shard decodes to ~921 MB and one assembled
   outer chunk is 5.98 GB (1440 samples x 721x1440 x f32). Measured on the fixed spike:
   **8.88 GB peak RSS at `resident_peak=1`**, 1440 samples in 6.9 s (~870 MB/s) — not slow,
   just fat. Unverified: what happens when a single tile exceeds `cache_budget_bytes`. The
   only over-budget message I can find lives on the `readonly_cache` path
   (`pool.py:844-851`), so the normal path may fail opaquely or deadlock. Worth one
   experiment, and worth a line in `docs/tuning.md` telling users to compute
   `spc x field x dtype` before choosing a budget.
3. **Docs**, either way: make `architecture.md:235-239` true, or stop claiming it.

Whole-shard vs partial-shard reads is a separate question — filed separately as an
optimization, not needed for correctness.

## Two adjacent defects found the same way

- **`open_geometries(store)` with no `variables=` crashes on any CF store**:
  `ValueError: sample_axis 0 out of range for 0-D array 'spatial_ref'`. Every dynamical
  dataset carries that 0-D grid-mapping scalar, as does most xarray-written zarr — and the
  README quickstart is exactly `open_geometries(store)`. Passing `variables=[...]` works
  around it; the error does not suggest it. Either skip non-indexable arrays or say what to do.
- **No documented route to a plain Icechunk repo.** `arraylake_store` is our only Icechunk
  entry point and it authenticates through Arraylake. A public Icechunk repo needs the
  `icechunk.s3_storage(..., anonymous=True)` -> `Repository.open` -> `readonly_session().store`
  form used in the reproducer above, which is a zarr-v3 `Store` the engine accepts and which
  appears in no README, doc page or example. "One contract, any backend" is load-bearing, so
  it should sit with the other store constructors.

emfdavid:
## Summary

Closes #56. Also fixes the two defects the same store surfaced: #57 is *not* closed by this
(the subset case is still open and linked from the docs).

On a sharded zarr-v3 array two shapes differ and only one is the storage unit:
`metadata.chunks` is the **inner** chunk — the read granularity *inside* a shard — while
`chunk_grid.chunk_shape` is the shard, which is what a chunk key addresses and what
`store.get` returns. Geometry, the read plan and the `ArraySpec` all used the former, so we
asked the store for a key holding a shard and decoded it as though it were one inner chunk.
The `ShardingCodec` then sized its index from the wrong spec, read the wrong trailing bytes,
and raised a checksum error that says nothing about sharding — which is why this presented as
"CRC32C mismatch" and cost an afternoon to attribute. Every dynamical.org dataset is sharded;
WB2 and ARCO are not, which is why every benchmark, example and test missed it.

The fix is one spelling of "the shape of one stored object" (`_storage_chunks`), used at the
three sites that had rolled their own. It reads through `ChunkGrid.from_metadata`, not
`metadata.chunk_grid` — the latter is deprecated on `ArrayV2Metadata`, and v2 is not optional
for us (ARCO is v2).

Two more, both found by the same store:

- **`open_geometries(store)` crashed on any CF group.** A 0-D `spatial_ref` grid mapping has
  no sample axis. Quieter and worse: when it didn't crash it returned `time`, `latitude` and
  `longitude` as variables to batch. Coordinates and grid mappings are skipped now, read from
  `dimension_names` (v3) or `_ARRAY_DIMENSIONS` (v2). The inference is deliberately narrow —
  only a **self-named** 1-D array is a coordinate, so a station series over `('time',)` stays
  data — and when the result is empty the error names every array it skipped and spells out
  the `variables=` escape hatch. There is a test asserting that string is in the message, so
  the escape hatch can't rot out of the error.

- **`cache_budget_bytes` below the residency floor was silently raised to the floor.** It now
  raises, naming the floor, the knob and the `block_chunks` that set it. Being handed ten
  times the memory you asked for is indistinguishable from being honoured right up until the
  kernel intervenes, and one chunk of a deep-sample-axis archive is 5.98 GB, not a rounding
  error. `None` still auto-sizes as before.

- **`icechunk_store(url)`** opens an Icechunk repository by URL — `s3://`, `gs://` or
  `file://`, anonymous or credentialed, `branch=` to pick the snapshot. `arraylake_store`
  stays for Arraylake-hosted repos; they are not two ways to do one thing, since Arraylake
  addresses repos by catalog name through its own client and no URL round-trips to one. The
  module docstring states that rule and the unknown-scheme error names the other constructor.

## For reviewers

**The behaviour change worth a second look is the budget raise, not the shard fix.**
`tests/test_residency.py::test_starvation_names_concurrent_iterations_as_the_cause` passed
`cache_budget_bytes=2 * chunk_bytes` with the comment "fits one iteration, not two" — but
that value was itself *under* the one-iteration floor, and the test had been quietly relying
on the promotion it was documenting the absence of. It now sizes automatically, which is what
it meant. That is a test changed alongside the behaviour it covers; please satisfy yourself
that the new form still tests starvation.

Also worth review: the coordinate inference is a heuristic on other people's metadata, and
heuristics on other people's metadata are where this kind of change goes wrong. I chose the
narrowest rule I could defend (0-D, or 1-D named for its own dimension) precisely so that the
failure mode is "you must pass `variables=`", never "we silently dropped your data". Both v3
and v2 spellings are parametrized in `tests/test_cf_coordinates.py`.

I am confident in the shard fix itself: `tests/test_sharded.py` builds a local store whose
shard shape differs from its inner chunk shape on purpose, and it reproduces the original
CRC32C failure with **no network** — so this defect is now caught offline, by anyone, in 0.4s.

Live-bucket coverage is marked `remote` and skipped unless `--remote` is passed. A public
store disappearing should not redden a build for someone who changed nothing; the reasoning
is in `tests/conftest.py` next to the marker.

## Checklist

      this change
      `uv run pytest -q` are green locally — 462 passed, 7 skipped; `--remote` 11 passed
      against the NOAA GFS analysis Icechunk store
- [x] User-facing behavior documented in `docs/*.md` — `architecture.md` (the sharding claim
      is now true, and links #57 for the subset case), `tuning.md` (new "Work out what one
      chunk costs before setting a budget", with the 5.98 GB worked example), `README.md`
      (constructor list)
- [x] A bullet added under `## Unreleased` in `CHANGELOG.md` — four
      (3.13t) run passes — CI will confirm; the change is to which *shape* the planner reads,
      not to concurrency
- [ ] Performance claim? — none made here

## Author attestation

Left unchecked by the assistant that drafted this: the box asserts a *human* reviewed every
change and verified the claims above, which is precisely the thing an assistant ticking it
would defeat. It is yours to tick.

https://claude.ai/code/session_012QLod5Wvcm4JKiCRGTBFdj

### DIFF
--- a/docs/tuning.md
+++ b/docs/tuning.md
@@ -272,6 +272,24 @@ A load that drops entries rewrites the log (to a temp name, then a rename), so a
 not re-read and re-rejected on every subsequent open. That is safe only because one writer
 holds the lock above.
 
+### Work out what one chunk costs before setting a budget
+
+`cache_budget_bytes` below what one pass needs co-resident now **raises at construction**,
+naming the floor, rather than being quietly raised to it. Compute the floor the same way the
+engine does — per variable, `samples-per-chunk × field × itemsize`, times `block_chunks`,
+times two (the current block plus one read-ahead):
+
+```python
+geom = open_geometries(store, variables=["temperature_2m"])["temperature_2m"]
+per_chunk = geom.sample_chunk_size * math.prod(geom.inner_shape) * geom.dtype.itemsize
+```
+
+This is not a rounding detail on an archive that chunks the sample axis deeply. A store
+holding 1440 time steps per chunk over a 721×1440 field costs **5.98 GB for one chunk** —
+so `block_chunks=4` asks for ~48 GB, and a box that cannot supply it should say so before
+the first read, not be OOM-killed halfway through an epoch. Passing no budget still sizes
+itself automatically.
+
 ### Put `cache_dir` on local NVMe, not NFS
 
 The cache is an mmap tier, so this is what it is built for. Over a network filesystem it
--- a/docs/tuning.md
+++ b/docs/tuning.md
@@ -274,21 +274,35 @@ holds the lock above.
 
 ### Work out what one chunk costs before setting a budget
 
-`cache_budget_bytes` below what one pass needs co-resident now **raises at construction**,
-naming the floor, rather than being quietly raised to it. Compute the floor the same way the
-engine does — per variable, `samples-per-chunk × field × itemsize`, times `block_chunks`,
-times two (the current block plus one read-ahead):
+`cache_budget_bytes` below what one pass needs co-resident **raises at construction**,
+naming the floor, rather than being quietly raised to it. Ask the geometry what one chunk
+costs, then check that against the memory you actually have:
 
 ```python
 geom = open_geometries(store, variables=["temperature_2m"])["temperature_2m"]
-per_chunk = geom.sample_chunk_size * math.prod(geom.inner_shape) * geom.dtype.itemsize
+geom.chunk_bytes                      # bytes one sample-axis chunk occupies once resident
+2 * block_chunks * geom.chunk_bytes   # the floor: current block + one read-ahead
 ```
 
-This is not a rounding detail on an archive that chunks the sample axis deeply. A store
-holding 1440 time steps per chunk over a 721×1440 field costs **5.98 GB for one chunk** —
-so `block_chunks=4` asks for ~48 GB, and a box that cannot supply it should say so before
-the first read, not be OOM-killed halfway through an epoch. Passing no budget still sizes
-itself automatically.
+Compare that with `free -h` (the `available` column, not `free`) and leave room for the
+model and the framework's own allocator. Across several variables the floor is the sum, and
+a `chunk_transform` can make the transformed output the binding term instead of the stored
+tiles — so for anything but a single-variable run, ask
+[`describe()`](api.md#insitubatch.InSituDataset.describe), which reports the number the
+engine will actually use rather than one you assembled by hand.
+
+**Do not compute this as `sample_chunk_size × prod(inner_shape) × itemsize`.** Residency is
+the array's stored *tiles*, kept whole: a grid that does not divide the array evenly still
+stores full-size edge chunks and the loader does not clip them. On NOAA GFS analysis —
+1440 time steps per chunk over a 721×1440 field, tiled 400×400 — that product gives 5.98 GB
+while the real cost is **7.37 GB**, eight whole tiles. Sizing a budget from the smaller
+number under-provisions by 19%, and the pool starves mid-epoch, which fails in the shape of
+a hang rather than an error.
+
+The difference is not a rounding detail on an archive that chunks the sample axis deeply.
+One chunk of that store is 7.37 GB, so `block_chunks=4` asks for **59 GB** — and a box that
+cannot supply it should say so before the first read, not be OOM-killed halfway through an
+epoch. Passing no budget still sizes itself automatically.
 
 ### Put `cache_dir` on local NVMe, not NFS
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -49,11 +49,13 @@ def run_by():
 def pytest_addoption(parser):
     """``--remote`` opts into the tests that read somebody else's live bucket.
 
-    Off by default: the suite is otherwise fully offline and deterministic, and a public
-    store going away must not turn into a red build for a contributor who changed nothing.
-    These tests exist because a local fixture cannot reproduce a backend-specific defect --
-    the sharded-decode bug (#56) lived in an Icechunk store and no synthetic fixture would
-    have found it.
+    They exist because a local fixture cannot reproduce a backend-specific defect: a
+    synthetic store exercises our code against our own writer, not against the encoding
+    choices a real publisher made.
+
+    Off by default, because the suite is otherwise fully offline and deterministic and
+    somebody else's public store going away must not turn into a red build for a
+    contributor who changed nothing.
     """
     parser.addoption(
         "--remote",
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -46,6 +46,32 @@ def run_by():
     return _run_by
 
 
+def pytest_addoption(parser):
+    """``--remote`` opts into the tests that read somebody else's live bucket.
+
+    Off by default: the suite is otherwise fully offline and deterministic, and a public
+    store going away must not turn into a red build for a contributor who changed nothing.
+    These tests exist because a local fixture cannot reproduce a backend-specific defect --
+    the sharded-decode bug (#56) lived in an Icechunk store and no synthetic fixture would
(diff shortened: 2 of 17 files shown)


docs/tuning.md
```diff
@@ -272,6 +272,24 @@ A load that drops entries rewrites the log (to a temp name, then a rename), so a
...
+### Work out what one chunk costs before setting a budget
+
+`cache_budget_bytes` below what one pass needs co-resident now **raises at construction**,
+naming the floor, rather than being quietly raised to it. Compute the floor the same way the
+engine does — per variable, `samples-per-chunk × field × itemsize`, times `block_chunks`,
+times two (the current block plus one read-ahead):
+
+```python
+geom = open_geometries(store, variables=["temperature_2m"])["temperature_2m"]
+per_chunk = geom.sample_chunk_size * math.prod(geom.inner_shape) * geom.dtype.itemsize
```
emfdavid: Make the per_chunk memory use a method (property?) on ArrayGeometry to make it easy for users.
What is the pattern to document - check this property, check your `free -h`... something like that?
Our shuffle strategy relies on using at least two chunks to approximate global shuffle, while a unshuffled usage only needs one chunk per preload right?

tests/conftest.py
```diff
@@ -46,6 +46,32 @@ def go() -> None:
     return _run_by
 
 
+def pytest_addoption(parser):
+    """``--remote`` opts into the tests that read somebody else's live bucket.
+
+    Off by default: the suite is otherwise fully offline and deterministic, and a public
+    store going away must not turn into a red build for a contributor who changed nothing.
+    These tests exist because a local fixture cannot reproduce a backend-specific defect --
+    the sharded-decode bug (#56) lived in an Icechunk store and no synthetic fixture would
```
emfdavid: Keep the reason it exists not how we found it

emfdavid (commented):
Looks good - two comments to respond to.

PR #59, opened 2026-09-06
docs: two working rules — validate the instrument, and keep living docs current

emfdavid:
## Summary

Two rules for `CLAUDE.md`, both from work this week where their absence cost real time.

**1. "Validate the instrument before the subject."** (new bullet under Working principles)

We already require a control/NULL run for performance claims. This is the same discipline one
step earlier — applied to the tool rather than the treatment. The motivating case: a zarr store
wrapper counting chunk reads returned **zero**, which reads as "no redundant reads at all" —
precisely the flattering answer I was half-expecting. Nothing raised. The cause was that opening
a store read-only *copies* it, so the counter stayed on an instance nobody held. Had I not
checked, I'd have published a confidently wrong number in the direction that suited us.

The defence costs nothing: point the instrument at a case whose answer you already know (16
samples out of one chunk must read 16) and require that answer first.

**2. "Docs: what goes where."** (replaces the `## Status / roadmap` section)

Living docs — README, `docs/*.md`, docstrings — describe the code as it is. `DESIGN.md` and
`CHANGELOG.md` own how it got that way. It absorbs the old Status/roadmap section rather than
sitting next to it, because "don't mirror milestone state, it goes stale" is the same rule with a
narrower scope; PEP 20 says put it in one place.

The test is mechanical, which is what makes it usable in review: **"used to", "previously", "no
longer", "instead of", "we considered" are each a signal the sentence belongs in `DESIGN.md` or
`CHANGELOG.md`.** Rationale for a *rejected* alternative goes to `DESIGN.md`; rationale for the
*implemented* pattern stays with the code, stated positively.

This is the rule #54 applied by hand to three docstrings. Writing it down is what stops it being
re-litigated per review.

## For reviewers

**Two things I decided rather than assumed, both worth overruling if you disagree:**

**No CHANGELOG bullet**, against the letter of our own "every PR adds a bullet" rule. Precedent is
on my side — `5b59838` and `dd6db9f` are docs+CLAUDE.md changes with no bullet — and CHANGELOG
documents user-facing behavior per release, which working notes are not. But it *is* an
inconsistency in the rule as written, and this PR edits the file that carries it. Worth settling:
either the rule should say "every PR that changes behavior", or I should add the bullet.

**I did not touch `docs/contributing.md`.** You scoped this to the Claude instructions, so I
stayed there. Flagging the gap rather than closing it: rule 2 binds anyone writing docs, and a
contributor adding "this previously worked differently" to `architecture.md` would be breaching a
rule they cannot see. The invariants list already carries an explicit pairing mandate with
contributing.md; this section does not, by choice, but it may deserve one. Say the word and it's a
three-line follow-up.

On wording: I refined both from your phrasing, and the substance is yours — the parts I added are
the *reason* each rule exists (a miswired instrument returns plausible numbers, not errors; a
living doc is read by someone deciding what to do next) since a rule without its reason is the
kind that gets argued with rather than followed.

## Checklist

- [x] User-facing behavior documented in `docs/*.md` — n/a, working notes only
- [ ] A bullet added under `## Unreleased` in `CHANGELOG.md` — deliberately omitted, see above
- [x] No load-bearing invariant is broken — no code changed; `mypy` passed via pre-commit, `ruff`
      had no files to check

## Author attestation

Left unchecked by the assistant that drafted this: the box asserts a *human* reviewed every
change and verified the claims above, which is precisely the thing an assistant ticking it
would defeat. It is yours to tick.

https://claude.ai/code/session_012QLod5Wvcm4JKiCRGTBFdj

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -13,6 +13,15 @@ Contributor-facing rules (scope limits, dev setup, AI policy) live in
   when you find a second way to do something, remove one.
 - **TDD where practical**, and *always* for bugs: first write the failing test
   that reproduces the bug, then fix until green. New behavior ships with tests.
+- **Validate the instrument before the subject.** Point any new measurement — a
+  counter, a probe, a benchmark harness — at a case whose answer you already know
+  and require it to produce that answer, *before* trusting it on the case you
+  don't. A miswired instrument almost never raises; it returns a plausible number,
+  and often a flattering one. (Opening a zarr store read-only *copies* it, so a
+  read-counter kept on the instance you hold counts nothing and reports "no
+  redundant reads at all" — indistinguishable from a real, favorable result.)
+  This is the same discipline as the control/NULL run required of performance
+  claims, applied one step earlier: to the tool, not the treatment.
 - **Fail fast.** Do not catch-and-continue on errors that cannot be genuinely
   recovered — let them propagate with context. Validate at boundaries and raise
   early. Use explicit exceptions for runtime contracts; reserve `assert` for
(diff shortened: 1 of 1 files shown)


emfdavid (commented):
:shipit:

PR #60, opened 2026-09-08
fix: release a pass's pool references after the scheduler stops, not before

emfdavid:
## Summary

Breaking out of a loader loop early — a smoke run over the first N batches, an evaluation
subset, `itertools.islice` — permanently burned part of the chunk pool, and a later epoch
then died with `residency budget exhausted ... this would hang`.

A pass released its pool references *inside* the scheduler's `with` block, i.e. **before**
the scheduler was closed. Closing is what stops the driver, and on a borrowed event loop it
is not synchronous: a cancelled-but-not-yet-unwound driver could still reach `try_admit` and
re-pin a slot under the owner that had just been released. Nothing releases that pin, and the
slot stays `FILLING` (its sibling tiles were cancelled) so it never becomes `READY` and
therefore never becomes evictable. The loss was cumulative and silent — throughput, shapes and
memory all look normal — until a budget sized for exactly two blocks had too little left to
admit the chunk a consumer was waiting for.

Diagnostic at the moment of failure, on `era5_c1` over S3:

```
active owners tracked: []          <- both passes released
pinned refs per owner: {1: 16}     <- the abandoned pass still holds 16 pins
slot states: {'FILLING': 16}       <- never READY, never evictable
```

It bit hardest where a batch pins a whole block (one sample per chunk,
`batch_size == block_chunks`) — the GRIB regime, and the benchmark suite's own `c1`
configuration, **every** run of which raised on S3 and GCS alike. The published `c1` number
was measured before the defect existed.

**Bisected to `5b8fae4`**, which did *not* introduce the over-pinning: it removed the
over-broad `unpin_all()` that had been quietly wiping it at every epoch boundary (#34). The
published revision `b594f8b` runs the same configuration fine; `cb0944c^` fails identically,
so #41 is not implicated.

The fix moves `release_owner` into the outer `finally`, after the `with` has closed the
scheduler.

## For reviewers

- **The ordering argument is the whole change** — worth a second look at whether
  `Scheduler.close()` is genuinely the last point at which a `try_admit` can occur.
- **A related defect is NOT fixed here and produces the same error message.** A fresh pass
  can still starve intermittently when a second pass shares the loop: the driver admits
  read-ahead until the budget is full, leaving no slot for the chunk the consumer is waiting
  for (in every sighting, the chunk being admitted is one nobody is waiting for, and `no tile
  is in flight`). Measured at ~5% (1/20) with a concurrent abandoned pass, 0/65 without one;
  doubling the budget to 64 chunks did not help, which is why I read it as demand-vs-read-ahead
  ordering (the **M-RA** roadmap item) rather than sizing. I'd like to file it separately
  rather than widen this PR.
- `tests/test_abandoned_pass.py` asserts the *invariant* (nothing is left held), not the
  consequence. A version asserting the consequence was written, **confirmed to pass against
  the broken code**, and deleted — the cloud-scale leak is too small to exhaust a
  CI-sized fixture. The three that remain fail against `main` (one with the production
  `RuntimeError`); the module docstring records this.

## Author attestation

**Not ticked — this was drafted by Claude Code, and that box is the author's to tick.**

## Checklist

      this change
      `uv run pytest -q` are green locally (482 passed, 7 skipped on this branch alone;
      the 7 are the CUDA-gated `test_pinned.py` cases, which pass 21/21 on an L4)
      (3.13t) run passes — **not run.** This changes teardown ordering around the scheduler,
      so it is in scope for that gate and someone should run it before merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,23 @@
 
 ## Unreleased
 
+- **Breaking out of a loader loop early permanently burned part of the chunk pool, and a
+  later epoch then died with "residency budget exhausted ... this would hang".** A pass
+  released its references *inside* the scheduler's `with` block -- before the scheduler was
+  closed. Closing is what stops the driver, and on a borrowed event loop it is not
+  synchronous: a cancelled-but-not-yet-unwound driver could still reach `try_admit` and
+  re-pin a slot under the owner that had just been released. Nothing releases that pin, and
+  the slot stays `FILLING` (its sibling tiles were cancelled) so it never becomes `READY`
+  and therefore never becomes evictable. The loss was cumulative and silent -- throughput,
+  shapes and memory all look normal -- until a budget sized for exactly two blocks had too
+  little left to admit the chunk a consumer was waiting for. It bit hardest where a batch
+  pins a whole block (one sample per chunk, `batch_size == block_chunks`), which is the
+  GRIB regime and the benchmark suite's own `c1` configuration: **every `c1` run of the
+  suite raised**, on S3 and GCS alike, while the published `c1` number was measured before
+  the defect existed. `release_owner` now runs after the `with` has closed. Bisected to
+  `5b8fae4`, which did not introduce the over-pinning: it removed the over-broad
+  `unpin_all()` that had been quietly wiping it at every epoch boundary (#34).
+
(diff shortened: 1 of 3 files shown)


emfdavid (commented):
:shipit:

PR #61, opened 2026-09-08
fix: fail a benchmark run when the engine under test does not run

emfdavid:
> Rebased onto `main` now that #60 is merged (`7db4e09`); this is a single commit against
> `main`, scoped to the benchmark harness and its docs.

## Summary

Three defects found while re-running the benchmarks for the release, all in the harness
rather than the engine.

**1. A run whose *subject* failed exited 0 with a tidy table.** Every engine failure was
caught and skipped so one flaky baseline could not kill an hours-long S3 sweep — but that
made a run in which `insitu` never ran look exactly like a complete one. It is how the `c1`
starvation fixed in #60 reached a results file as *absence*: `wrote 44 rows`, exit 0, and the
engine under test in none of them. Skips are now summarised at the end, and a skip of the
engine under test raises — **after** every row is on disk, since re-running an S3 sweep to
recover the configs that did work is expensive. A failing baseline stays non-fatal.

**2. The suite accepted an invalid cache label.** That guard immediately failed
`tests/test_plot.py`, which asks for `caches=("none", "memory")` — `"memory"` is not an insitu
cache mode, so those 4 configs had been raising and being silently swallowed. Fixed to
`("none", "resident")`, and `caches` is now validated at the boundary rather than on first use.

**3. Six documented commands could not be pasted.** The suite dropped `--storage` when it
started deriving storage from the URL scheme; the S3 invocations in `docs/benchmarks.md` and
`bench/benchmark_plan.md` kept passing it, so the runbook for a benchmark costing an EC2 box
and hours of reads died on `unrecognized arguments` before the first byte. `probe_memory` still
has the flag and keeps it.

## For reviewers

- **The scope of the raise is the judgement call.** It is deliberately narrow — only
  `engine == "insitu"` — on the grounds that a missing baseline is a gap in the comparison
  while a missing subject is not a benchmark. `test_a_failing_baseline_does_not_fail_the_suite`
  is the control that pins that distinction; if you'd rather it be a flag, say so.
- `tests/test_bench_docs.py` checks command *shape*, not values: it drops `choices` from the
  parser because the docs legitimately pass shell variables into choice-valued flags
  (`--backend "$BK"`). Mutation-checked — putting `--storage s3` back makes it fail.
- Reaching into `parser._actions` is the one piece of private-API use here; I couldn't find a
  public way to relax value domains for a shape check.

## Author attestation

## Checklist

- [x] Tests added or updated
      `uv run pytest -q` are green locally (496 passed, 7 skipped)
- [x] User-facing behavior documented — the corrected commands are the docs change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,23 @@
   `5b8fae4`, which did not introduce the over-pinning: it removed the over-broad
   `unpin_all()` that had been quietly wiping it at every epoch boundary (#34).
 
+- **A benchmark run whose *subject* failed exited 0 with a tidy table.** Every engine
+  failure was caught and skipped so one flaky baseline could not kill an hours-long S3
+  sweep -- but that made a run in which `insitu` never ran indistinguishable from a
+  complete one, which is how the `c1` starvation above reached a results file as *absence*
+  rather than as a failure. Skips are now summarised at the end, and a skip of the engine
+  under test raises -- after every row is on disk, because re-running an S3 sweep to
+  recover the configs that did work is expensive. A failing baseline stays non-fatal: that
+  is a gap in the comparison, not a dead run.
+
+- **Six documented benchmark commands could not be pasted.** The suite dropped `--storage`
+  when it started deriving storage from the URL scheme, and the S3 invocations in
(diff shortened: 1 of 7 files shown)


emfdavid (commented):
:shipit:

PR #62, opened 2026-09-09
fix: a stall is only terminal if the waiters really cannot proceed

emfdavid:
## Summary

Two independent defects in `Scheduler._starvation`, both of which **abort a healthy run**
with `residency budget exhausted ... this would hang` and tell the operator to raise a
budget that was never the problem.

The detector proves an admission stall terminal from three facts. Two of them were weaker
claims than they looked.

**1. A registered waiter is not necessarily a blocked one.** `blocked_waiters()` returned
`list(self._waiting)` — every waiter — but a waiter registers *before* it tests its
condition and is removed only once its thread is rescheduled. So a waiter whose chunk was
already READY and pinned by its own owner counted as blocked: the very waiter about to
wake, gather and unpin, which is the thing that frees budget. Captured at the moment of
the raise, on GCS:

```
unfinished tile tasks: 0     inflight_now: 0
waiters (chunk, slot_state, pinned): [(4780, 'READY', True)]
waiters whose wait is ALREADY SATISFIABLE: 1 of 1
```

**2. A queued delivery counted as no delivery.** `_inflight_now` counts only tiles that
have *acquired* the `max_inflight` semaphore. A tile task created and waiting its turn
delivers just as surely, and at `max_inflight=1` almost every tile is queued rather than
running — so a single iteration on a public GCS store, with an automatically sized budget,
was killed while 28 tile tasks were outstanding and its one waiter sat on a chunk in state
FILLING that one of those 28 was about to fill:

```
inflight_now=0
tile tasks created but NOT finished: 28
waiters (chunk, slot_state): [(1966, 'FILLING')]
```

That same configuration now drains its epoch (4800 samples).

`wait_ready` and `blocked_waiters` now share one definition of "can this waiter proceed"
(`_wait_satisfied`), with `_waiting` keyed by `(key, owner)` because the condition is
per-owner; and "is a delivery coming" is read off the tile tasks (`_tiles`), kept separate
from `_tasks` because the two answer different questions — what `close()` must cancel
versus what is still on its way.

### How they were found, and what was ruled out

Defect 1 surfaced as an intermittent failure of the benchmark suite (~1 pass in 20) with a
second pass sharing the loop, and never on a quiet process. Three other explanations were
tested and eliminated first: chunks released before their last use (an ordering probe found
0 violations per pass across four configurations), pins surviving `release_owner`
(0 leaked across six abandoned passes on GCS), and — for that instance — pending tile tasks
(0 outstanding at fire time). Raising `cache_budget_bytes` to 64 chunks changed nothing,
which is what ruled out sizing.

Defect 2 was found afterwards by sweeping loader configurations against a real store: 26
local cases were clean, and `max_inflight=1` failed only on GCS.

## For reviewers

- **The risk in this change is the opposite of the bug**: if either premise now
  *under*-reports, a real deadlock hangs instead of raising — the silent wedge the detector
  exists to replace. Both new predicates therefore ship with controls that must keep
  failing: a waiter on an unadmitted chunk is still reported, another owner's pin still does
  not satisfy your wait (#35), and a stall with nothing outstanding still raises. The
  existing `tests/test_residency.py` case (two interleaved iterations on a one-iteration
  budget must raise, naming concurrent iterations) also still passes.
- **The tests target the premises, not the timing.** Neither defect reproduces on a local
  fixture — the fetch is too fast to leave anything queued, and the scheduling window is too
  narrow — so an integration test would be a coin flip that reads like coverage. Each test
  constructs the state directly and is mutation-checked: under the old semantics, 2 of the
  first set fail and 1 of the second, while every control still passes.
- `_wait_satisfied` deliberately mirrors the wait predicate exactly, including the "failed
  slot" and "poisoned pool" cases — a waiter that is about to *raise* is also not blocked on
  budget. Worth a check that I got that correspondence right.
- Verified separately on a real store: 26/26 local edge cases across the pool/scheduler
  surface (abandoned passes, concurrent iterations, windowed offsets, transforms, mmap
  cache, persist/revive, readonly miss, bad chunks, sharded, straddling and oversized
  batches), each under a hard deadline so a hang counts as a failure.

## Author attestation

## Checklist

      this change
      `uv run pytest -q` are green locally (503 passed, 7 skipped; run pre-rebase on
      identical content, since the rebase only dropped #61's commit)
      (3.13t) run passes — **not run.** This is squarely in scope for that gate (it changes
      cross-thread readiness bookkeeping) and should be run before merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,24 @@
 
 ## Unreleased
 
+- **A healthy run could be killed with "residency budget exhausted ... this would hang".**
+  `Scheduler._starvation` turns an admission stall into a raise rather than a hang, and it
+  proves the stall terminal from three facts, the third being "a consumer is blocked in
+  `wait_ready`, so it can never reach its next unpin and free budget". That fact was read
+  off the registration in `ChunkPool._waiting` -- but a waiter registers *before* it tests
+  its condition and is removed only once its thread is rescheduled, so a waiter whose chunk
+  was already READY and referenced by its own owner -- one about to wake, gather and unpin,
+  which is precisely the thing that frees budget -- counted as blocked. Caught in the act on
+  a real store: at the instant the detector fired there were zero tile tasks outstanding and
+  exactly one waiter, on a chunk in state READY pinned by its own owner. The run was
+  correct, the working set was the intended two blocks, and it was aborted anyway. It
+  presented as an intermittent failure of the benchmark suite (~1 pass in 20) with a second
+  pass sharing the loop, and never on a quiet process -- the shape of a scheduling race, and
+  unaffected by raising `cache_budget_bytes`, which is what ruled out sizing. `wait_ready`
+  and `blocked_waiters` now share one definition of "can this waiter proceed", and only a
+  waiter that genuinely cannot is reported. A real deadlock still raises: the detector polls,
+  and an unsatisfiable waiter stays unsatisfiable.
+
 - **Breaking out of a loader loop early permanently burned part of the chunk pool, and a
   later epoch then died with "residency budget exhausted ... this would hang".** A pass
   released its references *inside* the scheduler's `with` block -- before the scheduler was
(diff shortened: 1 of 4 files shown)


emfdavid (commented):
:shipit:

PR #63, opened 2026-09-09
docs: list every command CI enforces, in all three places that claim to

emfdavid:
## Summary

`docs/contributing.md` opens **"The three commands CI runs"** and lists three. CI runs
**five**: `ruff check`, `ruff format --check`, `mypy`, `pytest`, and — on every pull
request — `mkdocs build --strict`.

So a contributor can do exactly what the guide says, see green locally, and still fail CI.
That is not hypothetical: it happened on our first outside contribution (#32), on a
formatting diff that says nothing about the change and gives no hint which command was
missing. The guide, the PR checklist and `CLAUDE.md` now list the same five.

Two further gaps the audit turned up, both in the guide only:

- **`mkdocs build --strict` runs on every PR**, and `--strict` promotes a broken link to a
  build failure — so a docs-only change can fail CI while every other command is green. The
  recipe, and the rule that pages under `docs/` cannot relatively link root files, were in
  `CLAUDE.md` but nowhere a contributor reads.
- **Tests marked `remote` skip unless `--remote` is passed.** Someone watching them skip had
  no way to know that is a complete run, or that CI does not pass the flag either.

### Pre-commit needs nothing added

Checked, since it is the other place these can drift. The hooks already run `ruff check`,
`ruff format` and `mypy` — anyone who ran `pre-commit install` was covered, and the gap was
only ever in the written path. Every tracked `.py` lives under the four paths CI checks
(`src tests bench examples`), so the hooks and CI cannot disagree about coverage. `pytest`
and the docs build stay out deliberately: both are too slow to sit on every commit, and the
guide is the right place to name them.

## For reviewers

- The wording change is "three" → "four" in the testing section plus a fifth (the docs
  build) called out separately, because it needs `--extra docs` and only matters when
  `docs/` changed. If you would rather see one flat list of five, say so.
- `CLAUDE.md` now says the three files list the same set, so a future edit to one is a
  prompt to check the others.
- Verified by running all five on this branch: ruff check, ruff format --check, mypy,
  `pytest -q` (496 passed, 7 skipped), and `mkdocs build --strict` (clean).

## Author attestation

## Checklist

      locally (`uv run --extra docs mkdocs build --strict` too — this PR touches `docs/`)
- [x] User-facing behavior documented in `docs/*.md` — this PR *is* the documentation fix

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -26,8 +26,9 @@
 
 - [ ] Tests added or updated — for a **bug fix**, a test that reproduces it and failed before
       this change
-- [ ] `uv run ruff check src tests bench examples`, `uv run mypy src bench examples` and
-      `uv run pytest -q` are green locally
+- [ ] `uv run ruff check src tests bench examples`, `uv run ruff format --check src tests
+      bench examples`, `uv run mypy src bench examples` and `uv run pytest -q` are green
+      locally (`uv run --extra docs mkdocs build --strict` too, if you touched `docs/`)
 - [ ] Docstrings and API docs for any new or changed public surface
 - [ ] User-facing behavior documented in `docs/*.md`
 - [ ] A bullet added under `## Unreleased` in `CHANGELOG.md`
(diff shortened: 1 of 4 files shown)


emfdavid (commented):
:shipit:

emfdavid:
@icn5381  Fixed the contributing guide to avoid lint errors in CICD. Thank you for your work on #32

PR #65, opened 2026-09-09
feat: bound read-ahead by the consumer, not by the pool (M-RA)

### LINKED ISSUE #27: `window_factor` sizes residency by offset span, not by the offset set
Filed by a maintainer from the *Known limitations & defects* list in [DESIGN.md](https://github.com/emfdavid/insitubatch/blob/main/DESIGN.md) — hence a plain issue rather than the bug form.

### The defect

`source.py:226` sizes the shuffle/residency working set from the offset **span**:

```python
window_factor = 2 + (-(-span // spc0)) if windowed else 1   # 2 + ceil(span/spc)
```

where `span = max(offset) - min(offset)`. For a **sparse** offset set the span is a poor proxy for how many chunks an anchor actually needs. Forecast leads `{24h, 240h}` on a one-sample-per-chunk store are 2 offsets with a span of 216, giving `window_factor = 218` — against the **2** chunks per anchor genuinely required. The memory budget is over-reserved by roughly 100×.

### What it should do

Size the factor from the number of **distinct read chunks per anchor**, not from `min(offset) → max(offset)`.

### What this is *not*

**It does not over-read.** `plan.py:_read_chunks` expands each offset independently — 10 sparse leads × 8 inits is exactly 80 reads, not a 240-wide span. Correctness is unaffected, which is why DESIGN.md marks it **low priority**. Please don't "fix" the planner; it is already right.

### Why it suits an outside contributor

The fix is local to one expression, the correct behaviour is already specified, and the test is an assertion on arithmetic rather than a benchmark: construct a dataset with sparse offsets on a `spc=1` geometry and assert the computed `working_set` reflects the distinct chunks per anchor rather than the span. Worst case in practice is chunk-1 archives (GRIB-style, per-frame, per-spectrum) with wide sparse leads.

New to the project? Start with the [contributing guide](https://emfdavid.github.io/insitubatch/contributing/).

### LINKED ISSUE #64: Concurrent iterations deadlock at any budget: admission is bounded by the pool, not by the iteration's working set
`zip(ds.train, ds.val)` deadlocks on any budget short of the whole split, because admission
is bounded only by the pool's byte budget and never by the iteration's own working set. One
driver pins **the entire budget**, whatever it is, and the second iteration never acquires a
slot.

This refines #38, which closed on the conclusion that sizing the budget for N concurrent
iterations is the remedy. That holds only where the budget can approach the size of the
whole split, which is what its fixture did — its table ran on a **32-chunk store and found
two iterations OK at a 32-chunk budget**, i.e. a budget covering 100% of the data. On a real
archive the same experiment never succeeds.

## Measurement

Public GCS store `gs://insitubatch-bench-insitubatch/era5_c1.zarr` (in-region), read from a
GCP `n2-standard-8` (8 vCPU, us-central1-b). Geometry `(6000, 721, 1440)` f4, one sample per
chunk, `chunk_bytes` 4.15 MB, 6000 chunks (4800 in the train split ≈ **19.9 GB**).
`batch_size=8`, `block_chunks=2` (widened to 8 so a batch fits one block), so one
iteration's residency floor is `2 x 8 = 16` chunks ≈ 66 MB. Interleaved
`next(train); next(val)`, as `zip` does. Both fixes from #62 are in, so the raise reflects
the corrected premises — this is a genuine terminal state, not a false alarm.

| budget | resident | pins by owner | outcome |
|---|--:|---|---|
| 2x floor (32 chunks) | 32 | `{1: 32}` | deadlock |
| 3x floor (48 chunks) | 48 | `{1: 48}` | deadlock |
| 4x floor (64 chunks) | 64 | `{1: 64}` | deadlock |
| 8x floor (128 chunks) | 128 | `{1: 128}` | deadlock |
| 16x floor (256 chunks) | 256 | `{1: 256}` | deadlock |

Owner 2 holds **zero** slots in every row. Owner 1's own working set is 16 chunks and it
holds 256 at the largest budget: the claim tracks the budget, not the need. Extrapolating
the pattern, the first budget that could work is one that holds the whole train split —
~19.9 GB for a 66 MB working set.

## Root cause

`Scheduler._admit` parks only when `pool.try_admit` fails, and `try_admit` fails only when
the byte budget is full. Nothing bounds how far ahead of its own consumer a driver may
admit, so a driver walks its plan pinning chunk after chunk until it has claimed the pool.
With one iteration this is invisible-but-costly (it is the eager-fetch behaviour DESIGN's
**M-RA** entry already records, inflating cold TTFB ~7 s → ~50 s). With two, it is fatal:
the second iteration cannot admit, so its consumer blocks; a blocked consumer never
unpins; and the first iteration eventually needs a slot that nobody can free.

At the 16x budget the pool held **113 slots simultaneously FILLING** — read-ahead depth is
likewise bounded by nothing but the budget, which is the memory face of the same defect.

## Why it matters beyond the deadlock

- **The error message's advice is wrong for this case.** It says to raise
  `cache_budget_bytes` and adds "the budget must cover all N" — raising it enlarges the
  over-claim. An operator following the message will keep raising the budget and keep
  deadlocking, having spent the memory.
- **`tests/test_residency.py` pins that advice.** It asserts this raise happens *and*
  asserts the message names concurrent iterations, so the guidance is currently regression-
  tested. Worth deciding what the message should say before changing it.
- `zip(ds.train, ds.val)` is documented as supported (`pool.py`, `buffers.py`,
  `tests/test_pinned.py`), so this is a defect against an advertised configuration.

## Fix direction

**M-RA** as already scoped: a fetch-ahead semaphore released on the consumer's `unpin`,
bounding how far ahead of its consumer a driver may admit, independent of retention. That
bounds a single iteration's read-ahead (the TTFB win M-RA is scoped for) and makes the
concurrent case work at the documented floor rather than at whole-split budgets.

A narrower alternative, if M-RA is too big for 0.2.x: cap admission per owner at that
owner's working set (`2 x block_chunks` chunks), leaving the rest of the budget as cache.
That fixes the deadlock without touching the fetch pipeline, but it does not deliver the
cold-TTFB improvement, and the two should not be built twice.

Repro script: `two_owner_probe.py` (sweeps the budget, dumps pins per owner at each
failure) — happy to attach it or fold it into `bench/` if useful.

emfdavid:
Closes #64. Closes #27.

Three commits: the read-ahead bound (#64), then two residency-floor fixes (#27) that the
first one made measurable — with `read_ahead_bound` computing the plan's real live set,
the floor could finally be checked against what a pass actually needs.

## Summary

Admission parked only when the pool's byte budget was full — never when a driver ran far
ahead of its own consumer. So a driver walked its plan pinning chunk after chunk until it
held **the whole pool**, whatever size that was. Measured on a public GCS store, one owner's
pins against its own 16-chunk working set:

| budget | resident | pins by owner | outcome |
|---|--:|---|---|
| 2x floor (32 chunks) | 32 | `{1: 32}` | deadlock |
| 4x floor (64) | 64 | `{1: 64}` | deadlock |
| 16x floor (256) | 256 | `{1: 256}` | deadlock |

The second iteration acquired **zero** slots in every row, so `zip(ds.train, ds.val)` — a
documented configuration — could not run at any budget short of the whole split (19.9 GB,
for a 66 MB working set). Raising `cache_budget_bytes`, which the starvation error advises,
enlarged the over-claim.

A pass now takes one **fetch-ahead permit** per chunk, returned when its consumer releases
that chunk, so it runs at most its own working set ahead of itself. Every budget above now
completes.

### The bound is computed, not configured

`read_ahead_bound(block_keys, last_use)` = `max over b of (live(b) + |block_keys[b+1]|)`.
There is no value a caller could choose that beats it — below it the pass deadlocks against
its own limit, above it is the behaviour this PR removes, and the benchmarks page already
establishes read-ahead depth as a cold-start knob rather than a throughput one.

**`2 x block_chunks` is not the bound.** Three configurations each falsified a simpler rule,
and each was found by hanging the suite rather than by reasoning:

- a **windowed** variable's chunk feeds several blocks and is released only at the last of
  them, so its permit spans all of them;
- a variable chunked **finer than the reference grid** contributes several chunks per anchor
  chunk, so any per-variable multiple of `block_chunks` under-counts;
- the driver is additionally allowed one block of read-ahead beyond what is live.

Both terms therefore come from the block read-unions the consumer actually waits on.

## Measurement

`era5_c8` over in-region GCS, GCP `n2-standard-8` (8 vCPU, us-central1-b), chunk 33.2 MB,
budget 13.3 GB (400 chunks), `block_chunks=16` (32-chunk floor), `batch_size=32`, 40
batches. Control and treatment are the same binary in the same session, arms alternated
three times:

| arm | TTFB | throughput |
|---|--:|--:|
| unbounded (today) | 1295 / 10094 / 965 ms | 533 / 400 / 562 MB/s |
| bounded (this PR) | 665 / 621 / 650 ms | 1082 / 1093 / 1152 MB/s |

Every bounded run beats every unbounded run on both axes, and bounded is far more
predictable. The 10-second outlier is the eager whole-split fetch this PR removes.

## Sizing the residency floor (#27)

`working_set_bytes` derived one `window_factor` from `max(offset) - min(offset)` and applied
it to every variable, and the whole-split clamp charged every *view* rather than every
array. Measured on a 2048-chunk store (1,638 in the train split), against the live set the
plan actually needs:

| offsets | floor before | after | measured live set | after / measured |
|---|--:|--:|--:|--:|
| none | 64 | 64 | 64 | 1.0x |
| dense ±1 | 4,914 | 1,640 | 1,356 | 1.2x |
| lead {0,24} | 3,328 | 1,662 | 898 | 1.9x |
| sparse {0,24,240} | **46,464** | **1,878** | 1,280 | 1.5x |
| sparse {0,24,48,96,240} | **77,440** | **1,878** | 1,711 | 1.1x |

The old floor demanded **28x the whole dataset** for three forecast leads, and it was not
merely wasteful: an explicit `cache_budget_bytes` of twice the real need was *rejected* as
below the floor, so that configuration could not be run at all. Sparse leads are what
dynamical.org's forecast archives look like.

**No latency claim.** Bounded read-ahead (the first commit) already decoupled admission from
the budget, so an inflated floor no longer makes the driver fetch further ahead: on the same
store, a 9x difference in budget moved TTFB not at all — 1372 ms at 46,464 chunks, 1373 ms
at 5,040. What the floor fixes buy is a runnable configuration and an honest `describe()`.

**One caution while fixing it:** charging per array without also extending the clamp by the
window's reach *under*-provisioned the five-lead case (floor 1,638 against a live set of
1,711). An under-sized floor starves mid-epoch, which is the failure the floor exists to
prevent, so the reach is added once in chunks — the additive form, not the multiplicative
one that made the original span term wrong.

## For reviewers

- **The free-threaded job is the gate I most want run.** This adds cross-thread signalling:
  permits are acquired on the loop and released from the producer thread via
  `call_soon_threadsafe`. I have no 3.13t environment, so CI is the first FT run.
- **Permit accounting is the correctness core.** Cache hits take a permit too — a hit is
  pinned exactly like an admission and released by the same `unpin_block`, so exempting it
  would return permits never taken. Audited over a full pass: 64 taken / 64 returned, plain
  and windowed.
- `tests/test_read_ahead_bound.py` tests the pure function for each configuration that broke
  a simpler rule, plus a harness that drives a real pass with a deliberately halved bound and
  asserts it wedges (fixture sized so a correct run takes well under a second against a
  20-second wait). That harness is why `SchedulerConfig.read_ahead_chunks` remains settable.
- **One measurement, one geometry.** I would not put the throughput numbers on the
  benchmarks page until a second store shape agrees.
- The floor fixes are checked against the *measured* live set rather than against a
  hand-derived formula, and every case is now at or above it — the direction that matters,
  since under-provisioning is a hang and over-provisioning is only memory.
- #27 was written up as suited to an outside contributor. I took it because it turned out to
  block a sane budget on the forecast-lead shape and because #65 made the correct number
  measurable; the harder follow-up it gestures at (re-fetching the window spill instead of
  holding the split) is untouched and still open for someone.

## Author attestation

## Checklist

- [x] Tests added or updated
- [x] `ruff check`, `ruff format --check`, `mypy`, `pytest -q` green locally (514 passed,
      7 skipped)
      (3.13t) run passes — **not run locally; this PR is how it gets run.**
- [x] Performance claim carries hardware, store, geometry, config and alternating controls

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,7 @@ site/
 
 # local backlog scratch (not tracked)
 ISSUES.md
+
+# Working notes for a design in progress: superseded the moment the code lands, and the
+# rules in CLAUDE.md send the parts worth keeping to DESIGN.md or CHANGELOG.md instead.
+design-notes/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,25 @@
 
 ## Unreleased
 
+- **The windowed residency clamp charged every *view* of an array, not every array.** With
+  shuffle on, a windowed pass holds the split resident; that clamp multiplied the split by
+  the number of variables, but `t2m.shift(0/24/240)` are three views of one array and the
+  pool keys slots by `(path, chunk)` -- the planner already collapses them into one fetch.
+  Three leads therefore reserved three times the memory one of them needs. Counted once per
+  array, and extended by the window's reach in chunks so the clamp still covers the chunks a
+  lead reads beyond the split's edges: sizing to the split alone under-provisioned the
+  five-lead case (a floor of 1,638 against a measured live set of 1,711), and an under-sized
+  floor starves mid-epoch, which is the failure the floor exists to prevent. With the span
+  fix above, sparse forecast leads size to 1,878 chunk-equivalents where they cost 46,464 --
+  1.1-1.5x the measured requirement rather than 28x the whole dataset.
+
+  This costs no latency, and it is worth being precise about why: bounded read-ahead already
+  decoupled admission from the budget, so an inflated floor no longer makes the driver fetch
+  further ahead. Measured on the same store, a 9x difference in budget moved
+  time-to-first-batch not at all (1372 ms at 46,464 chunks, 1373 ms at 5,040). What it buys
+  is a configuration that can be *run*: an explicit budget covering the real need was
+  previously rejected outright, and `describe()` reported a number up to 45x too large.
+
 - **The residency floor was sized by the distance between offsets, so sparse forecast
   leads reserved memory nobody needed (#27).** `working_set_bytes` derived a single
   `window_factor` from `max(offset) - min(offset)` and applied it to every variable, so
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,29 @@
 
 ## Unreleased
 
+- **One iteration could claim the entire chunk pool, so `zip(ds.train, ds.val)` deadlocked
+  at any budget (#64).** Admission parked only when the pool's byte budget was full, never
+  when a driver ran far ahead of its own consumer, so a driver walked its plan pinning
+  chunks until it held the whole pool: measured on a public GCS store, one owner held 32,
+  48, 64, 128 and 256 slots at 2x through 16x its 16-chunk working set, while the second
+  iteration acquired *none*. Raising `cache_budget_bytes` — what the error message advised
+  — enlarged the over-claim; the first budget that could have worked was one holding the
+  whole 19.9 GB split for a 66 MB working set. A pass now takes one fetch-ahead permit per
+  chunk, returned when its consumer releases that chunk, so it may run at most its own
+  working set ahead. Concurrent iterations work at the documented floor, and a single
+  iteration stops fetching the resident set before its first batch: on `era5_c8` over
+  in-region GCS with a 13.3 GB budget (GCP n2-standard-8, `block_chunks=16`,
+  `batch_size=32`, arms alternated three times in one session) time-to-first-batch went
+  from 965–10094 ms to 621–665 ms and throughput from 400–562 MB/s to 1082–1152 MB/s.
+
+  The bound is computed, never configured (`read_ahead_bound`), because there is no value
+  a caller could choose that beats it: below it the pass deadlocks against its own limit,
+  above it is the behaviour just described. `2 x block_chunks` is *not* that bound — a
+  windowed variable's chunk feeds several blocks and is released only at the last of them,
+  and a variable chunked finer than the reference grid contributes several chunks per
+  anchor chunk. Each of those found the rule by hanging the suite, and each now has a test.
+
+
 - **A healthy run could be killed with "residency budget exhausted ... this would hang".**
   `Scheduler._starvation` turns an admission stall into a raise rather than a hang, and it
   proves the stall terminal from three facts, the third being "a consumer is blocked in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+- **A fetch-ahead permit deficit hung the pass with nothing to read.** Bounded read-ahead
+  takes one permit per chunk and gets it back from the consumer's `unpin_block`, so a
+  consumer that cannot reach its next unpin never returns one. That wait was unbounded:
+  every other wait in the scheduler polls and proves itself terminal before giving up, but
+  this one presented as three idle threads, no log line, and no traceback -- the exact
+  failure the starvation detector was built to replace, reintroduced beside it. It bit
+  while developing per-block release, where a release that hands back fewer permits than
+  were taken is an easy mistake to make and, until now, an undiagnosable one. The wait is
+  now bounded and terminal on the same two facts as an admission stall (nothing in flight,
+  a consumer blocked where it cannot unpin), and it names the bound, the chunk, and what
+  the consumer is waiting on.
+
 - **The windowed residency clamp charged every *view* of an array, not every array.** With
   shuffle on, a windowed pass holds the split resident; that clamp multiplied the split by
   the number of variables, but `t2m.shift(0/24/240)` are three views of one array and the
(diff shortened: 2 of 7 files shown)


emfdavid (commented):
:shipit:

PR #67, opened 2026-09-09
fix: an empty split yields no batches instead of raising from inside numpy

emfdavid:
> **Stacked on #65** — base is `spike/m-ra-bounded-read-ahead`, not `main`. The diff here is
> one commit; review or merge #65 first.

## Summary

Iterating an empty split raised from inside numpy:

```
train/val/test chunks: {'train': 16, 'val': 0, 'test': 0}
  ds.val:  ValueError: need at least one array to concatenate
  ds.test: ValueError: need at least one array to concatenate
```

`fractions=(1.0, 0.0, 0.0)` is an ordinary request — train on everything, hold nothing back
— and any store small enough that 10% rounds to zero chunks produces the same shape by
accident. Both draw orders reached `np.concatenate([])`, and the message names neither the
split nor the dataset.

An empty split now yields nothing, the way iterating an empty list does. This is deliberately
*not* fail-fast: the manifest recorded the split as empty because it was asked to, so there is
no contract here that cannot be honoured. Both builders return an empty `(0, 2)` order,
keeping the shape callers index.

## How it was found

By running **every example** at a shrunken size — the first thing anyone does when trying the
library out. Five of the twelve failed this way: all four advection variants and SDSS.

Two of them then divided by no samples inside their own evaluation loop. Those now say so in
their own words rather than numpy's:

```
RuntimeError: no batches to evaluate: the split this view covers is empty. A small store
rounds a 10% split to zero chunks -- raise --n-steps, or widen the split's fraction.
Reporting skill over no samples would be worse than stopping here.
```

Worth noting the loader was never at fault in those two: advection *trained* fine
(`epoch 0 train mse 0.7200`) and only evaluation had nothing to work with.

**All 12 examples now run**, verified at viable sizes: transforms, fit_scaler, wb2_dataloader,
wb2_xbatcher, wb2_arraylake, the four advection variants, microscopy (IoU 0.579 vs Otsu 0.386),
hubble, and sdss.

## For reviewers

- The judgement call is empty-yields-nothing versus raising. I took the first because an empty
  split is a legitimate configuration, and because a user who asked for no validation data
  should not have to catch an exception to express that. If you would rather `ds.val` raise
  when the split is empty, that is a one-line change and the tests say which behaviour is
  pinned.
- `_EMPTY_ORDER` is copied rather than shared, so a caller mutating a returned order cannot
  poison the next empty split.
- The example guards raise rather than returning a sentinel score: reporting skill or IoU over
  zero samples is worse than stopping, and these are teaching material.

## Author attestation

## Checklist

- [x] Tests added or updated — written first, all six failing against the bug
- [x] `ruff check`, `ruff format --check`, `mypy`, `pytest -q` green locally (525 passed, 7 skipped)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,16 @@
   a consumer blocked where it cannot unpin), and it names the bound, the chunk, and what
   the consumer is waiting on.
 
+- **Iterating an empty split raised `ValueError: need at least one array to concatenate`.**
+  `fractions=(1.0, 0.0, 0.0)` is an ordinary request -- train on everything, hold nothing
+  back -- and any store small enough that 10% rounds to zero chunks produces the same shape
+  by accident. Both draw orders then reached `np.concatenate([])`, and the error named
+  neither the split nor the dataset, only numpy. It is what the advection and SDSS examples
+  do the moment their synthetic store is shrunk, which is the first thing anyone tries. An
+  empty split now yields nothing, the way an empty list does: the manifest recorded it as
+  empty because it was asked to, so this is not a contract that cannot be honoured. The two
+  examples that then divided by no samples say so in their own words rather than in numpy's.
+
 - **The windowed residency clamp charged every *view* of an array, not every array.** With
   shuffle on, a windowed pass holds the split resident; that clamp multiplied the split by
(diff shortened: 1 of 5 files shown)


emfdavid (commented):
:shipit:

Issue #66, opened 2026-09-09
Windowed + shuffled residency: the ordering, not the window, holds the whole split


emfdavid:
Windowed residency is not caused by windows, and not by shuffle. It is caused by the two
together, and today's ordering makes the combination as bad as it can be.

## The 2x2

512-sample store, one sample per chunk, 410 train chunks, `block_chunks=8`, `batch_size=8`.
"Live set" is the peak number of chunks that must be co-resident; "lifetime" is how many
blocks a chunk stays resident between its first and last use.

| | no shuffle | shuffle |
|---|--:|--:|
| **no window** `[0]` | 16 chunks (4%) | 16 chunks (4%) |
| **window** `[0, 1]` | 18 chunks (4%) | **229 chunks (56%)** |

Same offsets, same batch size, same block size: shuffle turns an 18-chunk working set into
229. Mean chunk lifetime goes from 1.1 blocks to 18.0, with a maximum of 51 out of 52 --
some chunk is needed in the first block of the epoch and again in the last.

That is why `working_set_bytes` clamps a windowed shuffled pass to the whole train split.
The clamp is roughly honest: measured live sets are 46-56% of the split against a clamp of
100%.

## Why

`block_shuffled_order` permutes at **chunk** granularity, then slices blocks off that
permutation. A windowed anchor `c` reads chunks `c` and `c+lead`, and chunk `c+lead` is
*also* the anchor chunk of another draw -- so two anchors need the same chunk, and the
permutation puts them in uniformly random blocks. Since the driver fetches each chunk once
per epoch (decode-once), that chunk must stay resident across the gap between them.

Residency is therefore set by chunk **lifetime** under the permutation, not by how many
chunks a batch touches. It gets worse as leads multiply, because each additional lead adds
another pair of anchors that must both be alive:

| offsets | live set | % of split | mean lifetime |
|---|--:|--:|--:|
| `[0]` | 16 | 4% | 1.0 |
| `[0, 24]` | 223 | 54% | 16.7 |
| `[0, 24, 240]` | 190 | 46% | 7.7 |
| `[0, 24, 48, 72]` | **397** | **97%** | 27.1 |

A 24/48/72-hour multi-lead training set -- an ordinary forecasting setup -- holds
essentially the entire split resident.

## Exploratory: ordering along the lead chain

The anchors that share a chunk are not random. For a lead stride `s` they are exactly
[...]

PR #69, opened 2026-09-09
fix: carry shuffle-block bounds with the order instead of reconstructing them

### LINKED ISSUE #68: Shuffle-block membership is reconstructed from the draw order, and the reconstruction is wrong on any windowed view
`_partition_blocks` recovers shuffle-block membership by *reconstructing* it from the
emitted draw order. On any windowed view that reconstruction is wrong, and the block a
chunk is planned into is not the block whose rows read it.

## The evidence

256 samples, `spc=4`, `block_chunks=2`, `shuffle=True`, two views of one array
(`{"now": base, "next": base.shift(12)}`), `fractions=(1.0, 0.0, 0.0)`:

```
rows=244  distinct chunks in order=61  block_chunks=2
expected blocks=31   recovered blocks=77

blocks whose declared cids != the chunk ids of their own rows: 66 of 77
  block 11: rows cover chunks [3], but blocks[11].cids = [3, 24]
  block 12: rows cover chunks [1], but blocks[12].cids = [30, 46]
  block 13: rows cover chunks [3], but blocks[13].cids = [17, 35]
  block 14: rows cover chunks [1], but blocks[14].cids = [21, 28]
```

`blocks[i].cids` is not a superset, not a subset, and not a permutation of the chunks
`blocks[i]`'s own rows draw from. It is a different partition of the same chunks.

## Why

`block_shuffled_order` builds the order over *all* the split's chunks and groups them
`block_chunks` at a time. `_clip_to_valid` then drops the edge anchors whose lead runs off
the array (#25) — here 12 rows, and with them three chunks entirely.

`_partition_blocks` re-derives membership from what survives:

```python
appearance = cids[np.sort(first_pos)]              # surviving chunks, first-seen order
block_of[appearance] = np.arange(len(appearance)) // block_chunks
```

That repacks 61 surviving chunks into fresh pairs. The rows, though, are still laid out in
the *original* 32-block order. From the first dropped chunk onward the two groupings shift
against each other, so `block_per_row` oscillates and the run-boundary scan finds 77 runs
where there are 32 blocks — while `appearance[k*bc:(k+1)*bc]` keeps handing out pairs from
the repacked grouping. The row ranges come from one partition and the chunk ids from the
other.

The information needed to do this correctly is not in `order`: a two-chunk block that lost
one chunk to clipping is indistinguishable from a one-chunk block. The reconstruction cannot
be fixed in place — the block structure has to come from the builder that created it.

## Blast radius

**Today, benign but unearned.** Both consumers of `blocks` that matter for correctness —
`block_keys` (what to pin) and `release` — go through `_block_read_keys(order[rs:re])`, which
reads the *rows*, so they stay self-consistent whatever the cuts are. `ordered_chunks`
flattens every block's cids and so still covers the right chunk set in the right order, which
is why fetch priority survives. The read-ahead bound is computed from the same row-derived
key sets: measured on the case above, 38 from the recovered blocks against 37 from the true
ones — no under-provisioning, and on the safe side of it. Nothing here is a live defect.

**For #66's per-block release, fatal.** Release-and-re-read is exactly the change that stops
routing both sides through the rows: the driver has to scope its dedup to a block
(`build_stored_chunk_reads(..., groups=)`, planning from `blocks[i].cids`) so a chunk a later
block reads is admitted again, while the consumer still releases
`_block_read_keys(order[rs:re])`. Those are the two sets above — the ones that disagree on
66 of 77 blocks. Every disagreement is a permit taken and never returned, or returned and
never taken, and the pass wedges. That is what the 3 failing tests on `feat/unpin-and-reread`
are failing on, and it is why #66 cannot land on top of this.

Retention is what has been hiding it: a chunk pinned from first use to last use is always
found no matter which block claimed to plan it.

## The shape of the fix

Have the order builder emit the block structure it created, rather than have the consumer
infer it. `block_shuffled_order` knows the boundaries exactly — it lays them down — and
`sequential_order` has the trivial one. Returning `(order, block_row_bounds)` (or an
`Order` carrying both) lets clipping shrink a block's row range without destroying the
identity of the block, deletes `_partition_blocks`, and removes the second definition of
"which chunks does this block read": there would be one, computed once, used by planner and
consumer alike.

Reproducer: `design-notes/` is ignored; the script above is ~30 lines against
`_partition_blocks` and `_block_read_keys` and can go in `tests/` with the fix.

## Related

- #66 — windowed + shuffled residency; this is its prerequisite, not its follow-up.
- #25 — the silent edge-anchor drop is the mechanism that makes the reconstruction wrong.
  Fixing the observability there does not fix this; fixing this does not remove the need for
  the count.
- #56 (closed) — same family: the planner disagreeing with the rest of the system about what
  a read is. Third time.
- #40 — the naming pass should land the vocabulary this touches (block, anchor chunk, read
  chunk) rather than adding a third spelling.

emfdavid:
## Summary

Closes #68.

A shuffle-block's chunks were **reconstructed** from the emitted draw order. `_partition_blocks`
repacked the surviving chunks into fresh groups of `block_chunks` while the rows stayed laid out
in the grouping the shuffle actually laid down. The moment `valid_anchor_range` dropped a whole
chunk -- which any lead offset does -- the two shifted against each other:

```
256 samples, spc=4, block_chunks=2, offsets {0, 12}
  rows=244  chunks 64 -> 61 (61, 62, 63 dropped)
  expected blocks=31   recovered blocks=77
  blocks whose declared chunks != the chunks their own rows draw: 66 of 77

  block 11: rows cover chunks [3], but blocks[11].cids = [3, 24]
  block 12: rows cover chunks [1], but blocks[12].cids = [30, 46]
```

This cannot be fixed in place: a two-chunk block that lost a chunk is indistinguishable from a
one-chunk block, so the information is not in the order. The boundaries now come from the builder
that creates them. `block_shuffled_order` and `sequential_order` return a `DrawOrder` carrying
rows *and* bounds, and `DrawOrder.keep` narrows the block that lost rows instead of re-cutting
every block after it.

That collapses the two definitions of "the chunks this block reads" into one `_Block`, derived
once from the rows the builder assigned it: `chunk_ids` for the driver to plan, `read_keys` for
the consumer to pin and release.

**Nothing observable changes, and that is worth being precise about.** The planner expands each
anchor into every chunk that anchor reads, and retention pins a chunk to its last use, so the
misattribution altered no sample and -- measured -- delayed no fetch: no block was fetch-complete
later than its own anchors, 0 of 77 on the worst geometry against 0 of 32 on the unwindowed
control. It was fatal only to the *next* change. Per-block release (#66) plans from a block's
chunks and releases that block's read keys, so every disagreement above is a read-ahead permit
taken and never returned; that is what the parked spike on `feat/unpin-and-reread` fails on.

This is the third time the planner has disagreed with the rest of the system about what a read
is (#56 was the last).

## For reviewers

**The API change is the thing to look at.** `block_shuffled_order` and `sequential_order` are
public (exported since 0.1.0) and now return `DrawOrder` rather than a bare array. `.rows` is
exactly what they used to return, and `sequential_order` gains a required `block_chunks` so both
orders carry one block definition. No docs page referenced either function. If you would rather
these stayed array-returning, the alternative is a second pair of functions that also return
bounds -- which is the "two obvious ways" this PR exists to delete.

**Behaviour is meant to be unchanged**, and the tests are built so that claim is checkable rather
than asserted. `tests/test_block_structure.py` (first commit, red before the second) takes every
answer from an oracle *outside* the code under test, so unifying the two definitions does not
turn any assertion into `f(x) == f(x)`:

* a per-sample walk that inverts `samples_in_chunk` instead of dividing -- kept permanently
  independent by the O(chunks)-not-O(samples) invariant, which forbids the engine from ever
  converging on it;
* `chunk_permutation`, for the blocks the shuffle actually laid down;
* an explicit live-set simulation, against `read_ahead_bound`'s open/close arithmetic.

19 of 115 failed on `main` across 7 geometries and 3 properties; all 115 pass here. Two
assertions were deliberately *not* written because unification makes them vacuous, and one that
did go vacuous (a block's chunk ids equal its own rows' chunks) was replaced by the disjointness
property it was really protecting.

Three instrument checks fired before any of this was trusted: a hand-computed peak that was
wrong, an oracle comparing within-block chunk order that carries no meaning, and a fourth claim
-- that misattribution inverts fetch priority -- that was measured, found false, and withdrawn.

## Author attestation

## Checklist

- [x] Tests added or updated — the reproducer commit is separate and fails on `main`
- [x] `ruff check`, `ruff format --check`, `mypy`, `pytest -q` and `mkdocs build --strict` green
      locally (641 passed + `test_tf.py` 2 passed in its own process)
- [x] Docstrings for the new public surface (`DrawOrder`)
- [x] No load-bearing invariant is broken — the hot path stays O(chunks); the O(samples) walk
      lives only in the test, which is what makes it a usable oracle
- [x] Touches the scheduler / cross-thread readiness → 3.13t run passes (CI)

### DIFF
--- a/src/insitubatch/source.py
+++ b/src/insitubatch/source.py
@@ -33,7 +33,7 @@ import queue
 import threading
 import time
 from collections.abc import Iterator, Sequence
-from dataclasses import replace
+from dataclasses import dataclass, replace
 from typing import TextIO
 
 import numpy as np
@@ -43,7 +43,7 @@ from .buffers import HostAllocator
 from .pool import ChunkPool, output_geometry, validate_scopes
 from .runtime import Depths, PassStats, StatsCollector, format_pass
 from .scheduler import Scheduler, SchedulerConfig
-from .shuffle import block_shuffled_order, sequential_order
+from .shuffle import DrawOrder, block_shuffled_order, sequential_order
 from .split import SplitManifest, valid_anchor_range
 from .store import close_store, open_geometries
 from .summary import DatasetReport, describe, print_summary, working_set_bytes
@@ -58,28 +58,21 @@ logger = logging.getLogger(__name__)
 DEFAULT_MAX_INFLIGHT = 32
 
 
-def _partition_blocks(order: np.ndarray, block_chunks: int) -> list[tuple[int, int, np.ndarray]]:
-    """Split a draw ``order`` into shuffle-blocks: ``(row_start, row_stop, chunk_ids)``.
+@dataclass(eq=False, slots=True)
+class _Block:
+    """One shuffle-block: where its rows are, what it draws from, and what it reads.
 
-    ``block_shuffled_order`` shuffles samples *within* a block of ``block_chunks``
-    chunks and concatenates blocks in chunk-permutation order, so every block is a
-    contiguous row range over disjoint chunks. We recover the blocks from the chunks'
-    first-appearance order (vectorized: O(chunks) of Python, not O(samples)), which
-    is robust to short final chunks where fixed-stride slicing would misalign.
+    Both halves come from the same rows. ``chunk_ids`` are the anchor chunks the driver
+    plans and fetches; ``read_keys`` are the ``(path, chunk)`` slots the consumer waits on
+    and releases, which a windowed view can push outside ``chunk_ids`` entirely. Deriving
+    them together, once, is the point: they were two derivations from two different inputs,
+    and on any windowed view they disagreed (#68).
     """
-    if not len(order):
-        return []
-    cids = order[:, 0].astype(np.int64)
-    _, first_pos = np.unique(cids, return_index=True)
-    appearance = cids[np.sort(first_pos)]  # chunk ids in order of first appearance
-    block_of = np.full(int(cids.max()) + 1, -1, dtype=np.int64)
-    block_of[appearance] = np.arange(len(appearance)) // block_chunks
-    block_per_row = block_of[cids]
-    starts = [0, *(np.flatnonzero(np.diff(block_per_row) != 0) + 1).tolist(), len(order)]
-    return [
-        (starts[k], starts[k + 1], appearance[k * block_chunks : (k + 1) * block_chunks])
-        for k in range(len(starts) - 1)
-    ]
+
+    start: int
+    stop: int
+    chunk_ids: np.ndarray
+    read_keys: set[tuple[str, int]]
 
 
 def read_ahead_bound(
@@ -391,7 +384,7 @@ class InSituDataset:
             ids = self.manifest.chunks[split.value]
         return np.asarray(ids, dtype=np.int64)
 
-    def _draw_order(self, split: SplitName | None, shuffle: bool) -> np.ndarray:
+    def _draw_order(self, split: SplitName | None, shuffle: bool) -> DrawOrder:
         chunk_ids = self._chunk_ids(split)
         spc = self._ref_spc  # the manifest's anchor grid, shared by every variable
         n_samples = self.manifest.n_samples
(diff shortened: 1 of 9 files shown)


emfdavid:
Validated on `insitubatch-bench2-b` (GCP, 8 vCPU, 31 GB, NVMe), 48-cell residency matrix over `gs://insitubatch-bench-insitubatch` (era5 at chunk depth 1/8/32 x offsets {} / {0,1} / {0,24} / {0,24,240} x single|zip x auto|4x budget, 12 batches per cell, 240 s deadline).

Measured against a baseline with #70 applied, so the permit-stall false positive is not in the way:

| comparison | outcome changes | residency changes |
|---|---|---|
| #70+#69 vs #70 alone | 1 | 8 |
| **#70 alone, run twice (control)** | **1** | **7** |

The change is indistinguishable from re-running the same code. Same cells, same magnitudes, and the peaks oscillate across repeats (465/470/480, 62/64, 88/89) — all of them `zip`, where two passes interleave on 8 vCPUs.

**No single-iteration cell moved in any comparison** — same outcome, same floor, same peak in all 24, every time.

Alongside the bit-identical batch A/B over 14 geometries and the 115 invariant tests, that is three independent lines all saying the same thing: the block structure was wrong, and nothing observable depended on it.

One incidental finding: `era5_c32 / lead 24 / zip / auto` flips ok<->RAISED on identical code. It sits right at the boundary of the known limitation that an `auto` budget is sized for one iteration, so it is not new here, but it makes that cell a poor regression signal.

src/insitubatch/source.py
```diff
@@ -58,28 +58,21 @@
...
-    ``block_shuffled_order`` shuffles samples *within* a block of ``block_chunks``
-    chunks and concatenates blocks in chunk-permutation order, so every block is a
-    contiguous row range over disjoint chunks. We recover the blocks from the chunks'
-    first-appearance order (vectorized: O(chunks) of Python, not O(samples)), which
-    is robust to short final chunks where fixed-stride slicing would misalign.
+    Both halves come from the same rows. ``chunk_ids`` are the anchor chunks the driver
+    plans and fetches; ``read_keys`` are the ``(path, chunk)`` slots the consumer waits on
+    and releases, which a windowed view can push outside ``chunk_ids`` entirely. Deriving
+    them together, once, is the point: they were two derivations from two different inputs,
+    and on any windowed view they disagreed (#68).
```
emfdavid: Drop the retrospective view

emfdavid (commented):
:shipit:

src/insitubatch/source.py
emfdavid: Dropped in 895729a, and the same framing in two neighbours I wrote: `DrawOrder` and `DrawOrder.keep` each explained the design by contrast with the reconstruction it replaces.

All four now state the current rule positively. The one thing worth keeping — why the bounds cannot be recovered from the rows after an edge-anchor drop — is in the CHANGELOG entry rather than lost.

641 passed, mypy clean.

PR #70, opened 2026-09-09
fix: a permit stall is this pass's own consumer, not another iteration's

emfdavid:
## Summary

A regression I introduced in #65, found while validating #68 on the GCS residency matrix.

`_ahead_starvation` proves a fetch-ahead permit stall terminal from two facts: nothing is on
its way to a slot, and a consumer is blocked where it cannot reach the unpin that would hand a
permit back. It read the second from `pool.blocked_waiters()` — **across every owner**.

That is right for `_starvation`, which judges the shared byte budget: another iteration's
blocked consumer really is holding budget we need, and that detector has a third fact besides
(`try_admit` just failed) to make the stall terminal. A fetch-ahead permit is not shared.
`_ahead` belongs to one `Scheduler` and comes back only from that pass's own `unpin_block`, so
another pass being parked in `wait_ready` says nothing about whether ours can proceed.

`zip(ds.train, ds.val)` interleaves two passes by construction, so at any instant one is very
likely mid-wait while the other's driver wants a permit. That ordinary overlap read as terminal:
the pass that was running fine raised `read-ahead permits exhausted` because the *other* one
happened to be waiting.

Same shape as the false positives in #62 — a detector asserting something false about the
future — reintroduced next door to the detector that fixed them.

## Measurement

48-cell residency matrix (`store x offsets x single|zip x budget`) on
`gs://insitubatch-bench-insitubatch`, `insitubatch-bench2-b` (GCP, 8 vCPU, 31 GB), era5 stores
at chunk depth 1 / 8 / 32, offsets `{}`, `{0,1}`, `{0,24}`, `{0,24,240}`, 12 batches per cell,
240 s deadline.

| run | ok | outcome changes vs the control |
|---|---|---|
| before #65's permit bound (control, same matrix, same box) | 42/48 | — |
| main today | 35/48 | 9 |
| **this branch** | **42/48** | **0** |

This branch reproduces the control cell for cell: same outcome in all 48, same floor in all 48,
and two small `zip` peak-residency deltas that are the interleaving of two passes, not a change
in what is held.

**8 cells went `ok` -> `RAISED`, every one a `zip` cell. No single-iteration cell moved**, in
outcome, floor, or peak. The remaining 6 failures are the pre-existing, honest ones: `zip` with
an `auto` budget is sized for one iteration and says so.

The zip failures were also *flaky* — re-running main against the #68 branch showed 3 cells
recovering and 2 newly failing — which is itself the signature: a race, not a budget.

## For reviewers

`blocked_waiters` now takes an optional `owner`; `_starvation` keeps reading every owner (that
is deliberate and its docstring says why), and `_ahead_starvation` passes its own.

The three new tests are predicate-level, using the white-box `_register_waiter` helper that
`test_starvation_detector.py` already established, so none of them depends on which thread the
OS runs next: a foreign owner's waiter must not be terminal (fails before this change), our own
must be, and a delivery in flight must not be. The second is the control that stops "never
report a permit stall" from passing the first and restoring the unbounded wait.

A slow consumer parked on queue backpressure registers no waiter at all, so it still correctly
reads as not-starved.

## Author attestation

## Checklist

- [x] Tests added — a failing predicate test plus two controls
- [x] `ruff check`, `ruff format --check`, `mypy`, `pytest -q` green locally (528 passed)
- [x] Touches `ChunkPool` / the scheduler / cross-thread readiness → 3.13t run passes (CI)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,19 @@
   array; its `.rows` is the array they used to return, and `sequential_order` takes
   `block_chunks` so both orders carry one block definition.
 
+- **The fetch-ahead permit stall judged itself by another iteration's blocked consumer, so
+  `zip(ds.train, ds.val)` started raising on runs that were fine.** Bounding that wait (a
+  fix earlier in this release) proved a stall terminal from two facts: nothing in flight,
+  and a consumer blocked where it cannot unpin. It read the second from the pool's waiters
+  across *every* owner -- which is right for an admission stall, because the byte budget is
+  shared and another pass's blocked consumer is holding some of it, and wrong for a permit,
+  which belongs to one scheduler and comes back only from that pass's own `unpin_block`.
+  Interleaving two passes makes that overlap the ordinary case rather than a rare one, so
+  the pass that was running fine reported `read-ahead permits exhausted` because the other
+  one happened to be mid-wait. Measured over a 48-cell residency matrix on GCS, 8 cells
+  went from `ok` to raising, every one of them a `zip` cell; no single-iteration cell was
+  affected. The permit stall now asks only about its own waiters.
+
 - **A fetch-ahead permit deficit hung the pass with nothing to read.** Bounded read-ahead
   takes one permit per chunk and gets it back from the consumer's `unpin_block`, so a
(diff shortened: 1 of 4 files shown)


emfdavid (commented):
:shipit:

PR #72, opened 2026-09-10
fix: import torch at runtime in as_torch, not only for type checking

emfdavid:
## Summary

`as_torch(view, device=...)` raises `NameError: name 'torch' is not defined` before a batch is
drawn. The branch deciding whether to page-lock the batch buffers names `torch.device(device)`:

```python
if device is not None and torch.device(device).type == "cuda":
    pin_host_buffers(stream, budget_bytes=pin_budget_bytes)
```

...but `frameworks.py` imports `torch` only under `TYPE_CHECKING`. Its other functions each
import it locally (lines 87, 102, 226); `as_torch` imports `IterableDataset` and not `torch`.

Any non-`None` device hits it, `"cpu"` included — the defect is naming `torch` at all, nothing
to do with CUDA:

```
device=None:  OK
device='cpu': NameError: name 'torch' is not defined
```

Present since `db35d94` ("track buffer liveness by data owner"), which added the `== "cuda"`
guard so a CPU device would not page-lock. Found running the GPU examples on an L4.

## Why nothing caught it

Every `as_torch` call in the tests, the README and `docs/index.md` omits `device=`, so the only
form exercised was the only form that worked. `docs/benchmarks.md` meanwhile points at
`as_torch(..., device=...)` as *the* route to pinned buffers — so the documented GPU path and
the tested path were disjoint, and the untested one was broken.

`tests/test_pinned.py` does cover page-locking, including 6 CUDA-gated tests that pass on an L4
— but it constructs `BatchBuffers(allocator=pinned_allocator())` directly. It tested the
mechanism thoroughly and never the entry point that reaches it.

## Verification on hardware

`insitubatch-gpu` (GCP, 16 vCPU, 62 GB, **NVIDIA L4 23 GB**, driver 580.173.02, torch
2.12.0+cu130). After the fix, iterating `as_torch(ds.train, device="cuda")`:

```
samples through as_torch(device=cuda): 512
peak CUDA bytes allocated            : 884736
batch buffers                        : BufferStats(kind='pinned', reuse=True, n_buffers=14,
                                                   nbytes=2064384, lends=102, allocations=14)
```

`kind='pinned'` is the part that matters: it is the allocator the `device=` branch exists to
install, and it was unreachable through the public API. Peak CUDA bytes > 0 is the positive
proof the device path executed rather than silently falling back.

Also on that box, on this branch merged with #69 and #70: `pytest -q tests/test_pinned.py` is
**21 passed, 0 skipped** (6 of those skip on a CPU box), and the full suite is 633 passed,
11 skipped (optional deps only).

## For reviewers

One line of source. The test passes `"cpu"` on purpose so it runs on every box rather than only
where a driver exists — a CUDA-gated regression test for this would skip in CI and on most
developer machines, which is how it got here.

Worth a second opinion on scope: `docs/benchmarks.md` documents a form no test exercised. Adding
a `device=`-carrying example to the README/docs is a separate call and I have not made it here.

## Author attestation

## Checklist

- [x] Tests added — a regression test that fails on `main`
- [x] `ruff check`, `ruff format --check`, `mypy`, `pytest -q` green locally
- [x] No load-bearing invariant is broken — torch stays an optional import inside the function

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,16 @@
   went from `ok` to raising, every one of them a `zip` cell; no single-iteration cell was
   affected. The permit stall now asks only about its own waiters.
 
+- **`as_torch(view, device=...)` raised `NameError: name 'torch' is not defined` before
+  drawing a batch.** The branch deciding whether to page-lock the batch buffers tests
+  `torch.device(device).type == "cuda"`, but `frameworks.py` imports `torch` only under
+  `TYPE_CHECKING` -- the other functions in the module each import it locally, and this one
+  did not. Any non-`None` device failed, `"cpu"` included; `device=None` was unaffected,
+  which is why it survived a month. Every `as_torch` call in the tests, the README and
+  `docs/index.md` omits `device=`, so the one form under test was the one form that worked,
+  while `docs/benchmarks.md` points at `device=` as the way to get pinned buffers. Present
+  since the buffer-liveness fix in v0.1.x; found running the GPU examples on an L4.
+
 - **A fetch-ahead permit deficit hung the pass with nothing to read.** Bounded read-ahead
   takes one permit per chunk and gets it back from the consumer's `unpin_block`, so a
   consumer that cannot reach its next unpin never returns one. That wait was unbounded:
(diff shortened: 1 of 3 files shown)


emfdavid (commented):
:shipit:

PR #73, opened 2026-09-10
feat: release a windowed shuffled pass's spill per block and re-read it

### LINKED ISSUE #66: Windowed + shuffled residency: the ordering, not the window, holds the whole split

emfdavid:
## Summary

Closes #66.

A windowed view reads `anchor + offset` and shuffle permutes chunk order, so a chunk one block
reads can be needed again most of an epoch later. Held from first use to last, residency was
close to the whole train split rather than two blocks.

With `persist=True` each block hands its chunks back as it drains, and a later block re-admits
them from the on-disk cache.

| offsets | policy | floor (chunks) | peak resident | re-reads | samples |
|---|---|---|---|---|---|
| `{0, 12}` | retain | 259 | 256 | 0 | 1012 |
| `{0, 12}` | **release** | **12** | **12** | 249 | 1012 |
| `{0, 24, 240}` | retain | 316 | 256 | 0 | 784 |
| `{0, 24, 240}` | **release** | **18** | **18** | 331 | 784 |

256-chunk split, `block_chunks=2`, two views of one array. Same samples delivered either way.

## Three parts, none of which works alone

* **plan** — dedup scoped to a block (`groups=`), so a chunk two blocks apart is admitted twice;
  the driver decides admission per chunk-run rather than once per pass.
* **consumer** — releases every key its block read, not only those at their last use; the
  read-ahead bound becomes three consecutive blocks (a batch on a boundary transiently holds the
  block behind it, the one it gathers, and the one it awaits).
* **floor** — `working_set_bytes` learns the policy. This one is easy to miss: a released chunk
  is only *evictable*, so with a floor still sized for retention nothing is ever evicted and
  releasing changes no number at all. That was the state of the parked spike.

## Two things measurement corrected

**The rule keys on `persist`, not `cache_dir`.** Only a persisted slot survives its own eviction
as a revivable file; with `cache_dir` alone the backing is unlinked on eviction, so the re-read
pays the fetch *and* the decode again — strictly worse than having retained the chunk. Same
fixture, same 249 re-reads:

```
persist=False   hits  10/501  (2%)   cache files   12
persist=True    hits 246/502  (49%)  cache files  256
```

The first draft of this keyed on `cache_dir` and logged "served from cache_dir", which would
have shipped a claim that was false in exactly the configuration it described.

**Why #66 was blocked, and no longer is.** The parked spike deadlocked because the driver planned
a block from its chunk ids while the consumer released the keys its rows read, and those
disagreed on 66 of 77 blocks. #68/#69 fixed that: block membership now comes from the builder, so
the planner's clamp and the consumer's edge-anchor drop agree. Verified before building on it —
across 11 geometries (plain, leads of 1/one-chunk/three-chunk, history+lead, sparse leads,
one-sample chunks, fat chunks, ragged tail, wide blocks, negative-only, a lead past the split):
**0 surplus permits, 0 shortfall, every block.**

## For reviewers

The `releases_spill` argument on `working_set_bytes` is the part worth a second look — it changes
a floor that raises at construction when a budget is below it, so getting it too low starves a
pass mid-epoch, which fails in the shape of a hang. It is three blocks rather than two for the
boundary reason above.

`docs/tuning.md` gains a section: what the trade is, why it is a rule and not a knob, and how to
read the churn line. The per-epoch summary reports re-reads and evictions whenever non-zero,
because a hit rate cannot show this — a re-read served from the cache counts as a hit.

Tests: 7 in `tests/test_reread_spill.py`, including the two controls that matter — an unwindowed
pass must not enable the policy (it would buy nothing and cost the dedup), and a retained pass
must report no re-reads (if it ever does, either the plan stopped deduplicating or a chunk is
being admitted twice for no reason).

## Author attestation

## Checklist

- [x] Tests added or updated
- [x] `ruff check`, `ruff format --check`, `mypy`, `pytest -q` (651 passed + tf's 2 in its own
      process) and `mkdocs build --strict` green locally
- [x] Docstrings for the changed surface (`working_set_bytes`, `Scheduler.start`,
      `build_stored_chunk_reads`)
- [ ] Touches `ChunkPool` / the scheduler / cross-thread readiness → 3.13t run passes (CI)

### DIFF
--- a/src/insitubatch/source.py
+++ b/src/insitubatch/source.py
@@ -259,6 +259,20 @@ class InSituDataset:
         # The last completed pass, or None before one finishes. Per pass, not per epoch: a
         # training epoch iterates train then val over one pool, and averaging two passes
         # with different shapes into one report is how a number stops meaning anything.
+        # A windowed shuffled pass holds a chunk from its first use to its last, which under
+        # a chunk permutation is most of an epoch -- so residency is close to the whole split
+        # (#66). Releasing each block's chunks when that block drains returns residency to the
+        # two-block floor, at the cost of admitting a chunk again when a later block reads it.
+        #
+        # A rule, not a knob, and it keys on `persist` rather than on `cache_dir`. Only a
+        # persisted slot survives its own eviction as a revivable file: with `cache_dir`
+        # alone the mmap backing is unlinked on eviction, so the second admission refetches
+        # and re-decodes. Measured on a 256-chunk windowed split, 249 re-reads served 2% from
+        # cache without `persist` and 49% with it -- the difference between the trade this
+        # policy assumes and the opposite one. Where the cheap path does not exist, retention
+        # is the better deal, so this turns on exactly where it does.
+        windowed = any(g.offset != 0 for g in self.geometries.values())
+        self.reread_spill = persist and windowed and shuffle
         self.last_pass: PassStats | None = None
         self._on_bad_chunk = on_bad_chunk
 
@@ -284,6 +298,7 @@ class InSituDataset:
             ref_spc=self._ref_spc,
             shuffle=self.shuffle,
             assembles=bool(self.chunk_transforms),
+            releases_spill=self.reread_spill,
         )
         # Sized for ONE iteration, deliberately. Every active iteration shares this pool and
         # holds its own references, so N concurrent iterations need ~N x this -- but the engine
@@ -496,11 +511,26 @@ class InSituDataset:
                 last_use[key] = bi  # later block overwrites -> ends on the max index
         for key, bi in last_use.items():
             release[bi].add(key)
+        if self.reread_spill:
+            # Every block hands back everything it read, so nothing is held for a later one:
+            # a chunk two blocks apart is admitted twice, and the pool serves the second from
+            # residency or from cache_dir.
+            release = [set(keys) for keys in block_keys]
 
         # Computed, not configured (see `read_ahead_bound`). A non-zero value on the
         # config is honoured so a test can supply a deliberately wrong one: a bound that
         # cannot be wrong is a bound nothing proves the need for.
-        ahead = self.scheduler_config.read_ahead_chunks or read_ahead_bound(block_keys, last_use)
+        if self.reread_spill:
+            # Three consecutive blocks, not the plan's live set. Nothing is held for a later
+            # block, so the long lifetimes that make the retained bound large do not arise --
+            # but a block is released only *after* the batch draining it has gathered, while
+            # the wait for the next block happens before, so a batch straddling a boundary
+            # transiently holds the block behind it, the one it gathers, and the one it awaits.
+            sizes = [len(keys) for keys in block_keys] or [1]
+            computed = max(sum(sizes[i : i + 3]) for i in range(max(len(sizes) - 2, 1)))
+        else:
+            computed = read_ahead_bound(block_keys, last_use)
+        ahead = self.scheduler_config.read_ahead_chunks or computed
 
         def produce(sched: Scheduler) -> None:
             # Batches are cut over the WHOLE epoch order, not per block. Cutting them per
@@ -520,7 +550,11 @@ class InSituDataset:
             bs = self.batch_size
             ready = freed = 0
             try:
-                sched.start(ordered_chunks, spc)
+                sched.start(
+                    ordered_chunks,
+                    spc,
+                    groups=[len(b.chunk_ids) for b in blocks] if self.reread_spill else None,
+                )
                 for start in range(0, len(rows), bs):
                     if stop.is_set():
                         return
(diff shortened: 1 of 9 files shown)


src/insitubatch/source.py
```diff
@@ -259,6 +259,20 @@ def __init__(
         # The last completed pass, or None before one finishes. Per pass, not per epoch: a
         # training epoch iterates train then val over one pool, and averaging two passes
         # with different shapes into one report is how a number stops meaning anything.
+        # A windowed shuffled pass holds a chunk from its first use to its last, which under
+        # a chunk permutation is most of an epoch -- so residency is close to the whole split
+        # (#66). Releasing each block's chunks when that block drains returns residency to the
```
emfdavid: Remove retrospective view

emfdavid (commented):
:shipit:

src/insitubatch/source.py
emfdavid: Dropped, and the same framing in three neighbours I wrote: `ChunkPool._note_touch`, the admission-decision comment in `_drive`, and `working_set_bytes`. Each explained the design by contrast with what it replaces, and the `reread_spill` one also carried the measurement that chose between the two.

All four now state the current rule. The measurement (249 re-reads served 2% from cache without `persist`, 49% with it) is in the CHANGELOG entry, which is where provenance belongs.

Swept the rest of the diff for the same pattern — three remaining hits on "would" are forward-looking statements about current behaviour, not retrospect.

652 passed, mypy and ruff clean.

PR #76, opened 2026-09-10
fix: answer an idle consumer before reading the batch queue

emfdavid:
## Summary

Fixes the `test-freethreaded` flake on #73, which turned out to be a real wrong diagnosis
rather than a fragile test.

`bottleneck()` read the batch queue before asking whether the caller's loop body did any work:

```python
if d.fed_frac >= FULL_ENOUGH:
    return "consumer", "...the loader kept up and your training step is the constraint..."
if stats.consumer_idle:
    # "with no training step the queue *cannot* stay fed, so its depth carries no signal"
```

That comment is the premise, and it is not always true. On a free-threaded build a
GIL-releasing `chunk_transform` runs alongside the consumer, so a `for batch in ds.train: pass`
loop can leave the queue looking fed — and then gets told its *training step* is the
constraint. CI reported exactly that, with a **19-microsecond** loop body:

```
StageTimes(assemble_s=0.371, decode_s=0.011, consumer_s=0.000019)
assert 'consumer' == 'decode'
```

Wall was 0.41 s. The advice sends someone to tune a knob they do not have, which is the
failure the rule's own docstring says costs more than an admission.

Whether the loop body did work is known before the queue is consulted, so it is asked first.

## Not just a flake

It re-ran green, and I could not reproduce it locally in 200+ attempts (isolation, whole file,
whole suite, free-threaded, 8 cores). That is the point: the verdict depended on how the
producer threads happened to interleave, so the same code gave different answers. The test
was right to fail; it just could not be relied on to.

The new tests construct `PassStats` directly, so neither the bug nor the fix depends on timing.
`test_a_fed_queue_does_not_blame_a_loop_body_that_does_nothing` fails on `main` with the same
`'consumer' == 'decode'`.

## For reviewers

The reorder makes the fed-queue verdict unreachable for an idle consumer, which is the whole
change — so the control test matters more than the fix:
`test_a_fed_queue_still_blames_a_consumer_that_is_doing_work` pins that a caller who is
genuinely busy still gets `consumer` and "the desired steady state". Without it, answering the
idle case first could quietly swallow the state the rule exists to recognise.

`docs/tuning.md`'s table had the same false premise ("your loop has no real training step, so
the queue cannot stay fed") and implied the wrong precedence. Its rows are now in precedence
order with that row first, and it no longer claims the queue cannot stay fed.

Verified on the build that flaked: `tests/test_runtime.py` green under 3.13t with
`PYTHON_GIL=0`, and the originally-failing test 10/10.

## Author attestation

## Checklist

- [x] Tests added — a failing test plus the control, both timing-free
- [x] `ruff check`, `ruff format --check`, `mypy`, `pytest -q` (655 passed) and
      `mkdocs build --strict` green locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 ## Unreleased
 
+- **A loop with no training step could be told its training step was the bottleneck.**
+  `bottleneck()` read the batch queue before asking whether the caller's loop body did any
+  work, so a `for batch in ds.train: pass` whose queue happened to look fed was answered with
+  "the loader kept up and your training step is the constraint" -- advice to tune a knob that
+  is not there. The `consumer_idle` branch below it was guarded by the premise that such a
+  loop *cannot* keep the queue fed; on a free-threaded build it can, because a GIL-releasing
+  `chunk_transform` runs alongside the consumer. It presented as a test that passed and
+  failed on identical code, with a 19-microsecond loop body reported as the constraint. The
+  idle case is known before the queue is consulted, so it is answered first, and the verdict
+  no longer depends on how the producer threads interleave.
+
 - **A windowed, shuffled pass held close to the whole train split resident (#66).** A windowed
   view reads `anchor + offset` and shuffle permutes chunk order, so a chunk one block reads can
   be needed again most of an epoch later; held from first use to last, residency was the split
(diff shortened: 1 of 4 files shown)


emfdavid (commented):
:shipit:

Issue #25, opened 2026-08-16
Expose a `dropped_edge_anchors` count on `InSituDataset`


emfdavid:
### What would you like insitubatch to do?

`InSituDataset` should report how many anchors it dropped at the edges of the sample axis, so a consumer can tell that its requested range silently shrank.

`source.py:_drop_edge_anchors` keeps only anchors whose every `anchor + offset` lands in `[0, n_samples)` and drops the rest with no counter and no signal. Dropping them is **correct and desired for training** — an incomplete window is not a usable sample — so the engine must not raise. What is missing is observability: an attribute such as `ds.dropped_edge_anchors` giving the count, or the effective-versus-requested anchor range.

### What does it unblock?

Any consumer that cares whether it got the range it asked for. Today the only way to detect shrinkage is to re-derive `valid_anchor_range` yourself and compare, so every consumer reimplements the same arithmetic. With wide or sparse windows on a short split the drop can be a large fraction of the epoch, and it is currently invisible — you get a smaller epoch and no indication why.

The inference/scoring path already closes this at its own boundary (the Earth2Studio adapter validates `sample_range` against `valid_anchor_range` and raises), which is the right place for a *policy*. This issue is only about making the fact observable, so any consumer can implement its own.

### Does this touch a load-bearing invariant?

No — this fits within the current contract.

### Scope

A counter set in `_drop_edge_anchors` and exposed on `InSituDataset`, a test that a window set which drops anchors reports a non-zero count and one that drops none reports zero, and a mention wherever the windowed path is documented.

**Deliberately out of scope:** an opt-in `on_edge="drop"|"raise"` knob. That is deferred as YAGNI until a second inference-style consumer appears — please don't add the knob for one caller. See the *Known limitations & defects* entry in [DESIGN.md](https://github.com/emfdavid/insitubatch/blob/main/DESIGN.md).

New to the project? Start with the [contributing guide](https://emfdavid.github.io/insitubatch/contributing/).

PR #77, opened 2026-09-10
fix: refuse an advection run that cannot be scored, at setup

### LINKED ISSUE #71: examples/advection: an --n-steps too small for a 10% split fails after training a full epoch, not at startup
`examples/advection` trains a full epoch before discovering that the split it is going to
evaluate on is empty. The check is correct and its message is actionable; it just runs far
too late, and the arithmetic that decides the answer is known before any work starts.

## Reproducing

```
uv run python -m examples.advection.train_torch --epochs 1 --n-steps 192
```

stdout:

```
make_advection_store: file:///tmp/insitu_advection.zarr
  3 vars x (192, 48, 48)  chunks=(48, 48, 48)  ~5 MB uncompressed  compress=auto
  wrote 192/192 steps (100%)  0.3s  748 steps/s
  done: file:///tmp/insitu_advection.zarr in 0.3s
epoch 0  train mse 0.8401
```

then exit 1:

```
RuntimeError: no batches to evaluate: the split this view covers is empty. A small store
rounds a 10% split to zero chunks -- raise --n-steps, or widen the split's fraction.
Reporting skill over no samples would be worse than stopping here.
```

The store is built, an epoch is trained, validation runs -- and only then does `evaluate()`
find the forecast view empty and stop. Nothing produced is kept.

## Why

`--n-steps 192` with `--sample-chunk 48` is 4 chunks. `split_by_chunk(fractions=(0.8, 0.1,
0.1))` gives the 10% splits `0.4` chunks, which rounds to zero. All of that is decided by the
arguments, before a byte is written.

The cliff is between 256 and 320 (default is 768):

| `--n-steps` | chunks | outcome |
|---|---|---|
| 192 | 4 | trains an epoch, then stops |
| 256 | 5 | trains an epoch, then stops |
| 320 | 6 | OK |
| 768 *(default)* | 16 | OK |

## Suggested fix

Validate at the boundary, where the house rule puts it: at argument parsing / dataset
construction, raise with the arithmetic rather than the symptom --

```
--n-steps 192 with --sample-chunk 48 is 4 chunks, and a 10% split of 4 chunks rounds to 0.
Use --n-steps >= 320, or --fractions.
```

Two things that would each be worth having on their own:

* it costs the user an epoch of training to learn a fact known at startup;
* the message names `--n-steps` but not the threshold, so the next guess is also a guess.

The engine-side half of this is already handled: an empty split yields no batches rather than
raising from inside numpy (#67). This is only about the example failing late.

## Notes

Found while validating #68/#69 on `insitubatch-bench2-b`. Not a regression -- the guard is
recent (#67) and before it the same configuration raised from `np.concatenate` instead, which
was worse. Filing separately because it is an examples/UX defect, not a loader one.

emfdavid:
## Summary

Closes #71.

`examples/advection` built its store, trained a full epoch, validated, and only then raised
from `evaluate` with nothing to score — discarding all of it to report a fact its arguments had
already fixed.

**The issue's premise was half right.** It said a 10% split rounds to zero chunks. Measuring
found two distinct causes with the same symptom:

| `--n-steps` | val chunks | drawable anchors | |
|---|---|---|---|
| 192 | **0** | 0 | the 10% split rounds away |
| 256 | 1 | **0** | the split is fine; the 24-step target drops every anchor in it |
| 288 | 1 | 8 | works |

At 256 nothing rounds to zero — the split lies entirely in the tail the window reaches past
(#25's silent edge-anchor drop). A chunk count cannot tell the two apart, which is why this
could not be fixed by counting chunks in the example.

## The diagnosis belongs in the engine

`describe()` now reports `drawable_samples` per split, and `print_summary()` shows it beside
the chunk count whenever the views are windowed:

```
splits (chunks)  train 5 (232 drawable)  val 1 (0 drawable)  test 0 (0 drawable)
```

A split holding chunks it cannot draw from is a `split-yields-nothing` warning naming the
offsets responsible. A split *asked* to be empty stays silent — warning on
`fractions=(1.0, 0.0, 0.0)` would fire on most of the examples and teach everyone to skip the
notes section.

The example reads those numbers rather than deriving its own, so there is one definition of
"can this split be drawn from".

## The advice is checked, not asserted

The refusal names the smallest sample-axis length that works, solved against the same
`split_by_chunk` and `drawable_samples` the run uses rather than quoted as a constant:

```
the val split has nothing to draw, so this run could train but never be scored. 256 samples
over chunks of 48 is 6 chunk(s); split (0.8, 0.1, 0.1) gives train=5 chunk(s)/232 drawable
samples, val=1 chunk(s)/0 drawable samples. A 24-step target also drops the last 24 anchors,
which can empty a split that did get chunks. Use at least 288 samples on the sample axis
(--n-steps for the synthetic store, --sample-range for a real one); you have 256.
```

`test_the_size_the_refusal_names_actually_works` builds a store at the named size and drains
its val split. A remedy nobody checked is how a clear error message becomes a second wrong
turn — and the number does adapt: `--sample-chunk 16` yields 128, which also works.

Corroboration worth noting: `tests/test_advection.py`'s fixture already used `n_steps=288`,
which is exactly what the solver returns for the defaults.

## For reviewers

The check runs after `InSituDataset` is constructed (it reads `describe()`) but before any
batch is drawn or any epoch trained — construction is pure geometry. It sits in
`forecast_dataset`, so every source path gets it, not just the synthetic one; for `wb2` and
`arraylake` the remedy is `--sample-range`, which the message says.

The controls are the interesting tests: an unwindowed run must have drawable == every sample
it holds, and a deliberately-empty split must produce no warning.

## Author attestation

## Checklist

- [x] Tests added — 4 on the report and its notes, 3 on the example
- [x] `ruff check`, `ruff format --check`, `mypy`, `pytest -q` (662 passed) and
      `mkdocs build --strict` green locally
- [x] Docstrings for the new public surface (`drawable_samples`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,23 @@
 
 ## Unreleased
 
+- **A windowed split could hold chunks and draw nothing, and only an empty iterator said so.**
+  A windowed view reads `anchor + offset`, so anchors whose window runs off the array are
+  dropped; a split lying entirely in that tail keeps its chunks and yields no batches. That is
+  a different cause from a split that rounded to zero chunks, with the same symptom, and the
+  chunk counts in `describe()` could not tell them apart. The report now carries
+  `drawable_samples` per split, `print_summary()` shows it beside the chunk count on windowed
+  runs, and a split with chunks but nothing to draw is a `split-yields-nothing` warning naming
+  the offsets responsible. A split asked to be empty is not a finding and is not warned about.
+
+- **`examples/advection` trained a whole epoch before finding it had nothing to score (#71).**
+  `--n-steps 192` on the default 10% split gives the val split no chunks; `--n-steps 256` gives
+  it a chunk whose every anchor the 24-step target drops. Either way the run built the store,
+  trained, validated, and only then raised from `evaluate` -- discarding all of it to report a
+  fact its arguments had already fixed. Both are now refused at setup, from the engine's own
+  `drawable_samples`, and the message names the smallest sample-axis length that works, solved
+  against the same split and window arithmetic rather than quoted as a constant.
+
 - **A loop with no training step could be told its training step was the bottleneck.**
   `bottleneck()` read the batch queue before asking whether the caller's loop body did any
   work, so a `for batch in ds.train: pass` whose queue happened to look fed was answered with
--- a/docs/tuning.md
+++ b/docs/tuning.md
@@ -293,6 +293,21 @@ tiles — so for anything but a single-variable run, ask
 [`describe()`](api.md#insitubatch.InSituDataset.describe), which reports the number the
 engine will actually use rather than one you assembled by hand.
 
+### A windowed split can hold chunks and still draw nothing
+
+A windowed view reads `anchor + offset`, so anchors whose window runs off the array are
+dropped. A split lying entirely in that tail keeps its chunks and yields no batches — the same
+symptom as a split that rounded to zero chunks, and a different cause. `describe()` reports
+both numbers, and `print_summary()` shows them side by side whenever the views are windowed:
+
+```
+splits (chunks)  train 5 (232 drawable)  val 1 (0 drawable)  test 0 (0 drawable)
+```
+
+A split with chunks but nothing drawable is a `split-yields-nothing` warning, naming the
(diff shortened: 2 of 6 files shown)


emfdavid (commented):
:shipit:

Issue #82, opened 2026-09-10
Plan the pass as a stream: bookkeeping is O(chunks) even where residency is O(block)


emfdavid:
Data residency is bounded and understood. The loader's *own* bookkeeping is not: every
structure built upfront is O(chunks) — one is O(samples) — with a Python-object constant, and
the whole pass is planned before the first fetch. On a store with O(10^9) chunks that is
hundreds of GB before a byte of data is read.

## Measured

Deep sizes with one shared identity set, so a structure is charged only for what it adds. Two
windowed views of one array, `block_chunks=4`, growth confirmed linear across 1k/4k/16k chunks:

| structure | B/chunk (spc=1) | B/chunk (spc=8) | grows with |
|---|---|---|---|
| `manifest.chunks` | 36 | 36 | chunks |
| `DrawOrder.rows` | 16 | 128 | **samples** (16 B/sample) |
| `DrawOrder.bounds` | 1 | 2 | blocks |
| `blocks` (`_Block`: cids + read_keys) | 299 | 422 | chunks |
| `ordered_chunks` | 37 | 37 | chunks |
| `last_use` | 40 | 43 | chunks |
| plan `reads` (`StoredChunkRead`) | 149 | 149 | chunks x inner tiles |
| **total (pass)** | **577** | **817** | |
| `_seen_this_pass` (pool) | 150 | 150 | chunks touched |
| `_persisted` + `_recorded` (pool, `persist=True`) | 325 | 325 | **chunks cached, not chunks read** |

Extrapolated: **577–817 GB at 10^9 chunks.** Practically, bookkeeping alone caps the design at
roughly **14M chunks per 8 GB**, or ~55M in 32 GB — one to two orders short of the target.

## Almost all of it is representation, not information

* `manifest.chunks["train"]` is a Python `list` of `int` holding a **contiguous ascending run**:
  118,036 B where the same run is 26,216 B as an `int64` array and **48 B as a `range`**.
* residency keys are `(path, chunk)` tuples: **84 B** each before set overhead, against **8 B**
  packed into an int64 with a per-path id.
* `plan.reads` materialises one `StoredChunkRead` per stored tile **for the entire epoch**,
  inside `Scheduler.start`, before the first fetch.

Re-expressed as arrays with packed keys, the same information is ~24–40 B/chunk — still 24–40 GB
at 10^9, so representation alone does not get there.

## The cross-run cache index is worse, because it outlives the pass

[...]

PR #83, opened 2026-09-10
fix: a pass reports its own counters, not the pool's

### LINKED ISSUE #81: Per-pass stats report pool-wide counters, so concurrent iterations overwrite each other's numbers
`PassStats` and the per-epoch summary read counters that live on the shared `ChunkPool`, so
with two concurrent iterations each pass reports the *pool's* activity under its own split's
name — and `reset_epoch_counters()` is pool-wide, so each pass zeroes the other's totals when
it starts.

## Measured

512 samples, `spc=4`, `fractions=(0.8, 0.1, 0.1)`, one variable:

| | hits | misses | resident peak |
|---|---|---|---|
| `ds.train` alone | 0 | **102** | **4** |
| `ds.val` alone | 0 | 13 | 4 |
| `zip(ds.train, ds.val)`, reported as `split='train'` | 0 | **29** | **35** |

Train alone misses 102 — every chunk it reads. The zipped run reports 29 for the same split,
because `ds.val`'s pass called `reset_epoch_counters()` partway through and threw away what
train had accumulated. The peak of 35 is neither pass's: it is how much the pool held for both.

## Which fields

Everything in this list is pool state read back as if it were the pass's:

* `cache_hits` / `cache_misses` (`pool.hits`, `pool.misses`)
* `depths.resident_peak` / `resident_peak_bytes` (`pool.max_resident*`)
* the churn line's `rereads` and `evictions`
* `_seen_this_pass`, which decides what counts as a re-read — one pass's take marks a chunk
  seen, so the other pass's first take of it is reported as a re-read

`Depths`' queue samples, the stage times and `inflight_peak` are *not* affected: those come from
a `StatsCollector` minted per pass.

## Why it matters

`last_pass` exists to say what to tune. Under `zip(ds.train, ds.val)` — a pattern the docs show
— it answers about a pool while naming a split, and a hit rate computed from two passes' reads
is not a hit rate for either. The residency numbers are the ones most likely to be acted on, and
they are the most distorted: 35 against 4.

Not a correctness bug: nothing about what is delivered depends on these.

## Same shape as #68, #70, #79

A fact that belongs to one owner, read off a structure shared by several:

* #68 — block membership recovered from the emitted draw order
* #70 — a permit stall judged by *any* owner's blocked waiter
* #75 — a fetch skipped because *some* writer holds the slot
* #79 — a chunk's admission run merged across a block boundary
* this — a pass's residency read from the pool's counters

The fix has the same shape too: scope the state to the owner whose report it is. The pool
already keys references by owner (`_pinned[key][owner]`), so per-owner counters are not a new
concept, only an unapplied one.

## Notes

Found while auditing for other instances of the #79 seam, after a 96-cell residency matrix over
GCS. The concurrent-iteration budget refusal added in #73 makes an under-sized `zip` fail at the
boundary now, so the *common* way to hit this is a `zip` with an explicit budget large enough to
run — which is exactly the configuration someone tuning from these numbers would be in.

emfdavid:
## Summary

Closes #81.

`PassStats` and the per-epoch line read hits, misses, residency peaks, re-reads and evictions
off the `ChunkPool`. Several iterations hold one pool, and `reset_epoch_counters()` ran at each
pass's **start** — so the second pass zeroed the first's totals mid-flight and then reported the
union under its own split's name.

| | hits | misses | residency peak |
|---|---|---|---|
| `ds.train` alone | 0 | **102** | **4** |
| `ds.val` alone | 0 | 13 | 4 |
| zipped, reported as `split='train'` — **before** | 0 | **29** | **35** |
| zipped, each pass — **after** | 0 | val 13, train 22 | **4** and **4** |

After the fix the val pass reports exactly what it reports when run alone (13 chunks, peak 4),
and the train figure is its partial run — the zip ends when val is exhausted. The pool's own
peak for the same run is 37, which is the pool's answer to a different question.

## What changed

Each iteration carries a `PassCounters`, minted with its owner in `new_owner` and dropped in
`release_owner`; the report reads those. Residency is counted where a reference is taken and
dropped, so it measures what *that owner* held rather than what the pool did — including the
awkward case where a consumer pins a key before it is allocated, which now charges its bytes to
everyone already holding it once the slot exists.

The pool keeps its totals. Peak co-residency and evictions genuinely belong to the pool, and
losing them would be a different mistake. But they are **cumulative** now: resetting them at one
pass's start is what made them mean nothing, and nothing in the reporting path reads them any
more, so `reset_epoch_counters` is left holding only the batch-buffer reset.

`_seen_this_pass` is deleted as *pool-wide* state — it decided a per-pass statistic, so one
iteration's take made another's first take look like a re-read. It is **not** gone as a cost:
the same set now lives on each `PassCounters`, measured at the same 150 B/chunk, and there is
one per live iteration rather than one per pool. Correct now, and strictly more memory (see
"Known cost" below).

## Known cost, not fixed here

`PassCounters.seen` is 150 B/chunk **per live iteration** — the same per-chunk set as before,
replicated rather than removed. It exists only to decide whether a take is a re-read, and it is
the single most expensive structure per chunk in #82's tally for something that is purely a
statistic.

It is derivable rather than discoverable: under a released spill a chunk is admitted once per
block that reads it, so re-reads over the drained blocks are
`sum(|block.read_keys|) - |union of them|` — a property of the plan and how far the pass got,
computable from structures that already exist. That would take the set to a plain int. Left out
of this PR because it is a separate change to a separate question, and this one is about
attribution; raised so the trade is on the record rather than discovered later.

## For reviewers

The residency accounting is the part worth a second look: `_pin`/`_unpin_one` maintain a live
count and byte total per owner, and the peak is taken on every increase. The 0→1 transition is
what counts, so a second reference to a chunk an owner already holds costs it nothing more,
which is the intended reading of "what this pass held".

Not fixed here, and worth knowing: the **batch-buffer** counters in the same epoch line
(`N lent, M allocated`) have exactly this shape — `BatchBuffers` is shared and its stats are
reset per pass. Scoping those needs the owner threaded through `gather`, which is a wider change
than this one; I left it and would rather it were its own issue than a silent half-fix here.

## Author attestation

## Checklist

- [x] Tests added — a concurrent-pass attribution test
- [x] `ruff check`, `ruff format --check`, `mypy`, `pytest -q` (659 passed) green locally
- [ ] Touches `ChunkPool` → 3.13t run passes (CI)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## Unreleased
 
+- **A pass's report described the pool, not the pass (#81).** Every counter behind `PassStats`
+  and the per-epoch line -- hits, misses, residency peak, re-reads, evictions -- lived on the
+  `ChunkPool`, which several iterations hold at once, and `reset_epoch_counters()` ran at each
+  pass's *start*. So with `zip(ds.train, ds.val)` the second pass zeroed the first's totals
+  mid-flight and then reported the union under its own split's name: a `val` pass whose true
+  residency peak is 4 reported 35, and a `train` split that misses 102 chunks alone reported 29.
+  Those are the numbers someone sizing a budget acts on.
+
+  Each iteration now carries its own `PassCounters`, minted with its owner and dropped with it,
+  and the report reads those. The pool keeps its totals -- a genuinely different question, and
+  the only one an eviction or a peak co-residency belongs to -- but they are cumulative now
+  rather than reset by whichever pass started last.
+
 - **A released spill could unpin a chunk out from under the block that still needed it (#79).**
   The driver decides admission once per *run* of tiles naming one chunk, since reads are
   chunk-major. Under release-and-re-read the plan emits a chunk once per block that reads it --
--- a/src/insitubatch/pool.py
+++ b/src/insitubatch/pool.py
@@ -464,15 +464,14 @@ class _TileWrite:
 
 @dataclass(slots=True)
 class PassCounters:
-    """One iteration's own view of the pool, so a report is about the pass that made it.
+    """One iteration's own view of the pool: what *this* pass hit, missed, held and evicted.
 
-    The pool's totals answer a different question -- what the *pool* did, across every
-    iteration sharing it -- and both are worth having. What is not worth having is one
-    labelled as the other: with two iterations live, pool totals attributed to a split are
-    the union of both passes' work under one pass's name (#81).
+    Minted with an owner and dropped with it, so the figures a pass reports are the figures it
+    made. The pool's totals answer the neighbouring question -- what the pool did, across every
+    iteration sharing it -- and both are worth having under their own names.
 
-    ``max_resident`` here is how many chunks this owner held at once, not how many the pool
-    did. Those differ precisely when it matters.
+    ``max_resident`` is how many chunks this owner held at once, not how many the pool did.
+    Several iterations share a pool, so those differ exactly when there is more than one.
     """
 
     hits: int = 0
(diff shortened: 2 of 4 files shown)


emfdavid (commented):
:shipit:

PR #85, opened 2026-09-10
fix: the batch-buffer figures are the pool's, and say so

### LINKED ISSUE #84: Batch-buffer counters are pool state under a per-pass label, and lends double-counts under zip
The batch-buffer segment of the per-epoch line is pool state printed under one split's name,
and one of its numbers is wrong when two iterations run at once.

## Measured

512 samples, `spc=4`, `fractions=(0.8, 0.1, 0.1)`, `batch_size=8`.

Sequential — the ordinary epoch — is correct throughout:

```
epoch 0 (train): ... batch buffers 3 x heap = 0.0 MiB, 51 lent, 3 allocated   (51 batches)
epoch 0 (val)  : ... batch buffers 3 x heap = 0.0 MiB,  7 lent, 0 allocated   ( 7 batches)
```

One lend per batch, and val's `0 allocated` is the pool correctly reusing what train left.

Interleaved (`zip(ds.train, ds.val)`):

```
epoch 0 (val)  : ... batch buffers 7 x heap = 0.1 MiB, 14 lent, 5 allocated   ( 7 batches)
epoch 0 (train): ... batch buffers 7 x heap = 0.1 MiB, 16 lent, 5 allocated   ( 8 batches)
```

A pass that drew **7 batches reports 14 lends**. Each pass reports the pool's running total at
its own finish, which under a 1:1 interleave lands at about twice its own — the doubling is
arithmetic, not a fixed factor.

## Field by field

* **`lends` — incorrect.** A per-pass quantity read from shared state. Same defect as the chunk
  counters in #81, and the same cause: `BatchBuffers` is shared and `reset_counters()` runs at
  each pass's *start*, so the second pass zeroes the first's count and then both accumulate
  together.
* **`allocations` — a pool fact under a per-pass label.** Its documented job is "allocations
  should fall to zero once the pool has converged; a nonzero count in a later epoch means
  buffers are not coming back". That is a statement about the pool and it survives intact. What
  breaks is the labelling: `5 allocated` on two lines reads as 10, and neither pass caused all
  five.
* **`n_buffers`, `nbytes`, `kind`, `reuse` — correct.** Genuine pool properties, rightly
  reported as such.

## Suggested direction: label it, do not scope it

#81 scoped the chunk counters per owner because they genuinely *are* per-pass — a pass's hit
rate and residency peak are facts about that pass. The buffer pool is different: four of its six
fields are pool properties already, and the fifth (`allocations`) is only useful as one.

So the cheaper and more accurate fix is to report that segment as what it is — the shared buffer
pool's state — rather than thread an owner through `gather` to attribute `lends`. That also
retires the per-pass `reset_counters()`, which is what corrupts it: the convergence diagnostic
wants a *cumulative* allocation count, not one reset by whichever pass started last.

If `lends` is worth keeping as a per-pass number at all, the thing it checks is "one lend per
batch", which `PassStats.batches` already gives for free.

## Notes

Found while fixing #81; deliberately left out of that PR rather than half-fixed. Severity is
lower than #81's: nobody sizes a budget from these, and the pool-level reading of `allocations`
still works. It is the same family as #68 / #70 / #75 / #79 / #81 — a fact belonging to one
owner, or to the pool, read off the other.

emfdavid:
## Summary

Closes #84.

One buffer pool serves every iteration open on a dataset, and its counters were reset at each
pass's **start** — so under `zip(ds.train, ds.val)`:

```
epoch 0 (val)  : ...  7 batches  ->  14 lent, 5 allocated
epoch 0 (train): ...  8 batches  ->  16 lent, 5 allocated
```

Each pass reports the pool's running total at its own finish. Sequentially the same run is
correct (`51 batches -> 51 lent`), which is why it survived.

## Labelled, not scoped — the opposite call from #81

#81 scoped the chunk counters per owner because a pass's hit rate and residency peak genuinely
*are* facts about that pass. These are not: `n_buffers`, `nbytes`, `kind` and `reuse` are pool
properties already, and `allocated` is only useful as one — its whole job is to say whether the
pool has converged. Threading an owner through `gather` would buy attribution for `lent`, and
`PassStats.batches` already gives what that checks.

So the segment is labelled `(pool)`, the per-pass reset goes away, and the counters run
cumulatively — the only form that means the same thing however many iterations share the pool.
`reset_epoch_counters` had nothing left to do and is deleted.

## For reviewers

The guidance changes: **"allocations fall to zero" becomes "allocations stop rising"**. Same
diagnostic, in the form a running total takes. `docs/tuning.md` quotes the line and is updated
with it.

Two test changes worth checking rather than skimming, both deliberate:

* `test_epoch_summary_reports_buffer_state` pinned the per-epoch reset (`lent0 == lent1 == 8`).
  It now pins the cumulative meaning (`8`, then `16`) and asserts convergence as a **delta**.
* its free-threading bound compared `held` against `lent1`. With `lent1` cumulative that bound
  slackens as a run gets longer, so it now compares against one epoch's batch count. That is a
  weakening I introduced and then had to take back out; flagging it because it is the kind of
  thing a semantics change does quietly.

## Author attestation

## Checklist

- [x] Tests updated to the new semantics
- [x] `ruff check`, `ruff format --check`, `mypy`, `pytest -q` (659 passed), `mkdocs --strict`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+- **The batch-buffer figures in the epoch line were the pool's, printed under one split's name
+  (#84).** One buffer pool serves every iteration open on a dataset, and its counters were reset
+  at each pass's *start* -- so with `zip(ds.train, ds.val)` a pass that drew 7 batches reported
+  14 lends, each pass reporting the pool's running total at its own finish. Unlike the chunk
+  counters (#81) these are not per-pass quantities to begin with: four of the six fields are
+  pool properties already, and `allocated` is only useful as one. So the line labels the segment
+  `(pool)` and the counters run cumulatively, which is the only form that means the same thing
+  however many iterations share it. The guidance changes with it -- watch `allocated` stop
+  rising rather than return to zero.
+
 - **A pass's report described the pool, not the pass (#81).** Every counter behind `PassStats`
   and the per-epoch line -- hits, misses, residency peak, re-reads, evictions -- lived on the
   `ChunkPool`, which several iterations hold at once, and `reset_epoch_counters()` ran at each
--- a/docs/tuning.md
+++ b/docs/tuning.md
@@ -153,13 +153,18 @@ give it back afterwards. If the burst fit, the floor fits.
 
 `close()` is the only release, and it drops the cross-epoch chunk cache and store session with
 it — so size the budget for your burst rather than planning to reclaim between phases. The
-per-epoch log line reports the pool directly, and `allocated` staying nonzero after the first
-epoch is the signal that batches are not coming back:
+per-epoch log line reports the pool directly, and `allocated` continuing to climb after the
+first epoch is the signal that batches are not coming back:
 
 ```
 epoch 3 (train): chunks 151/151 hit (100%), peak resident 51;
-                 batch buffers 17 x pinned = 544.0 MiB, 100 lent, 0 allocated
+                 batch buffers (pool) 17 x pinned = 544.0 MiB, 400 lent, 17 allocated
 ```
+
+The chunk figures are that pass's own; the buffer figures are the pool's running totals, marked
+`(pool)` because one buffer pool serves every iteration open on the dataset. Read `allocated` as
+a number that *settles* — 17 after four epochs of 100 batches each is a pool that converged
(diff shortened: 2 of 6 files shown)


emfdavid (commented):
:shipit:

PR #86, opened 2026-09-10
release: 0.2.0

emfdavid:
Cuts 0.2.0, plus one docs commit the cut itself taught us.

**`aa0025a` — the cut** (three lines):

- `pyproject.toml` — `version = "0.2.0"`
- `CHANGELOG.md` — `## Unreleased` becomes `## 0.2.0 — 2026-09-11`
- `uv.lock` — follows the version

**`5da23a2` — order the release process around the constraint that binds.** Doing this cut
surfaced that the checklist in `docs/contributing.md` listed the right steps in no particular
order, and was missing two outright: it never said to bump the version, and never said to
write the release notes. It also did not state the one ordering rule that actually matters —
the README is the only prose in the wheel and is what PyPI renders, so whatever it says at
bump time is what that version says on PyPI for good. The steps are now grouped into *before
the version bump* / *the bump* / *verify what you are about to ship* / *tag and announce*, so
the constraint is structural rather than a remark. The release-notes step carries the
compression rule and one trap: an issue reference that meant another project's tracker in the
changelog renders on a GitHub release page as *this* project's issue number.

> **Rewritten 2026-09-11.** This branch was force-pushed onto `main` after #90 landed, because
> the earlier cut conflicted on all four of its files. The previous head was `1c373ef`. Its
> README hunk is **deliberately not carried over**: it added a clause about `persist=True`
> releasing each block and re-reading from cache to a Status paragraph that no longer exists,
> and that behaviour now has its own account in the README's Caching section, stated more fully
> than the clause managed. The changelog grouping and `### Upgrading` preamble also landed in
> #90, so nothing is left for this PR but the cut.

## Verified against the release process, not assumed

Run through the checklist in `docs/contributing.md` (added by #90):

- **The three maturity statements agree** — classifier `Development Status :: 4 - Beta`,
  README `Status: Beta`, DESIGN.md `Maturity: Beta`. They were two steps apart before #90:
  the classifier still said `2 - Pre-Alpha`, which is what PyPI shows in the sidebar.
- **Five gates green** plus `mkdocs build --strict` — 668 passed, 9 skipped.
- **The artifact was opened, not assumed.** The wheel carries `py.typed`, the
  `insitubatch-check-transform` console script and LICENSE. The sdist's only prose is
  `README.md` — which is what PyPI renders — and METADATA embeds the rewritten one
  (22,090 bytes, with the stale "alpha, but validated" text confirmed gone), with
  `Project-URL: Changelog` pointing at the file.
- **The wheel installs into a clean venv**: `insitubatch.__version__ == "0.2.0"`, the public
  API imports (`InSituDataset`, `DrawOrder`, `to_jax`, `applies`, `icechunk_store`), and the
  console script runs.
- **GPU gate** cleared on an L4 earlier today: the CUDA-gated buffer-lifetime tests ran green
  with the instrument validated first (`CUDA_VISIBLE_DEVICES=""` skips exactly those tests),
  and persistence RMSE held at 0.884 across the CPU and CUDA legs.

Because the README is the only prose in the wheel, the rewrite had to land *before* this bump
or 0.2.0 would have shipped the page the docs-usability test failed on.

## For reviewers

The only judgement call was already made in #90 — that Beta is claimed on instrumentation
rather than on an absence of bugs. This PR just makes the version match.

## Author attestation

**Left unchecked deliberately** — that box is the human author's to tick.

## Checklist

- [x] `ruff check`, `ruff format --check`, `mypy`, `pytest -q` and `mkdocs build --strict`
      green locally
- [x] No load-bearing invariant is broken — no source change
- [x] A bullet under `## Unreleased` — n/a; this PR *is* the heading rename, and every entry
      it covers already has one

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## 0.2.0 — 2026-09-11
 
 ### Upgrading
 
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -295,52 +295,81 @@ decides — is written down in
 
 ## Cutting a release
 
-A maintainer task. Every step below has caught something at least once, and they are in order.
+A maintainer task. Every step below has caught something at least once, and the order matters
+in one place above all: **the docs are finished before the version is bumped.**
 
-1. **Group the changelog.** Contributors append flat bullets under `## Unreleased` — there is
+`README.md` is the only prose in the wheel, and it is what PyPI renders. Whatever it says when
+you bump the version is what that version says on PyPI for good — a README fix landed
+afterwards helps the *next* release, not the one people are installing.
+
+### Before the version bump
+
+1. **Run the docs-usability test.** Give someone with no prior context — a person or an agent
+   — an unfamiliar public store and *only* the published docs, and ask them to report the
+   store's geometry, build a working batch iterator, add a chunk-stage and a batch-stage
+   transform, and exercise the cache cold, warm, and from a second process. The deliverable is
+   the friction log: where they were stuck, what they guessed, what they believed that was
+   wrong, and which docs they wanted. "The docs were enough" is a legitimate finding. It is
+   the only test that measures the documentation rather than the code, and acting on it is
+   what this phase is for.
+
+2. **Execute every runnable block in the docs, rather than reading them.** Extract each one
+   and run it. Console transcripts should be regenerated from a real run of the configuration
+   the page builds, not edited by hand; a transcript carried over from an earlier
+   configuration is indistinguishable from a current one at the moment of reading. Check that
+   links resolve, and remember that a relative link in `README.md` resolves against
+   `pypi.org` and breaks — anything a reader must reach needs an absolute URL.
+
+3. **Group the changelog.** Contributors append flat bullets under `## Unreleased` — there is
    deliberately no category to choose while developing. The release cut is where they get
-   sorted under the `###` headings, by whoever is already in the file renaming the heading to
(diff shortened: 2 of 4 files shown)


emfdavid:
## Final integration — green

Run against `e8dd238` (main + #85), on `insitubatch-bench2-b` (GCP, 8 vCPU, NVMe).

**96-cell GCS residency matrix** (`gs://insitubatch-bench-insitubatch`, era5 at chunk depth 1/8/32 x offsets {} / {0,1} / {0,24} / {0,24,240} x single|zip x auto|4x budget x retain|persist), 12 batches per cell:

| | ok |
|---|---|
| before #80 | 67/96 |
| **this candidate** | **72/96** |

All 24 remaining failures are the *same* kind — `cache_budget_bytes cannot hold 2 concurrent iterations` — which is #73's intended boundary refusal, raised at construction with the budget to pass. **The 4 `read-ahead permits exhausted` cells are gone**, which is #80 confirmed on real cloud stores rather than only on the synthetic reproducer.

**Examples**, all 7 OK, with the model-independent fingerprints unchanged from every previous run this cycle:

| example | fingerprint |
|---|---|
| advection_torch | persistence RMSE **0.884** |
| adv_metrics | persistence RMSE **0.884** |
| microscopy | Otsu IoU **0.386** |
| sdss | PCA baseline 0.2027 (within its own ±0.0002 noise) |
| hubble, transforms, wb2_dataloader | OK |

Those are properties of the delivered bytes, so they are what would move if any of #69/#70/#72/#73/#76/#77/#80/#83/#85 had quietly changed what the loader returns. They have not moved all cycle.

Checklist item 'Final integration run attached' is satisfied.

emfdavid (commented):
🎉

PR #89, opened 2026-09-11
fix: place a JAX batch on JAX's default device, not the host

### LINKED ISSUE #88: JAX users on a GPU box silently train on the CPU
**Corrected 2026-09-11 after verifying on the L4 — the original framing of this issue was
wrong. See the comment below for what changed.**

## The finding, measured

On the GPU box (L4, `jax[cuda12]`, jax 0.11.1, `jax.devices() == [CudaDevice(id=0)]`), against
a local zarr fixture:

```
VALUES: exact vs store over 8 batches / 64 samples
DEVICE placement from to_jax: ['cpu:0']
device_put -> {CudaDevice(id=0)}   values intact: True
jit(x*2) result device: {CpuDevice(id=0)}      <-- the problem
```

`to_jax` is `jnp.from_dlpack(numpy_array)` (`src/insitubatch/frameworks.py:259`). That imports
a **host** buffer, so the result is committed to `cpu:0` — correct, and zero-copy, which is
what the docstring promises. The trap is the last line: **`jax.jit` does not move a
host-committed array to the GPU.** A user who follows the README on a GPU box gets a pipeline
that is correct, silent, and entirely on the CPU.

Values are exact throughout, so nothing is *wrong* — it is just slow, with nothing raised to
say so. That is the same shape as the defect class this project keeps finding: plausible
output, no error, and only a measurement tells you.

## Two separable things

1. **The device story is undocumented.** `device=` exists on `as_torch`/`to_torch` only;
   `to_jax` and `as_tf_dataset` take no device argument. Nothing said so. **Fixed in the
   README** as part of the 0.2.0 docs pass: device placement is called out as torch-only, and
   the JAX snippet now does an explicit `jax.device_put`, verified on the L4 to land on
   `cuda:0` and to keep a subsequent `jit` there. A `device_put` against a CPU-only JAX is a
   no-op, so the same line is correct on any box.
2. **The `jax` extra still resolves to the CPU wheel.** `jax = ["jax>=0.4.30", ...]` in
   `pyproject.toml` gives a GPU box a CPU JAX, so `uv sync --extra jax` on a CUDA machine
   silently cannot use the GPU at all. Still worth a `jax-cuda` extra pinning `jax[cuda12]`,
   with the README line pointing at it. That part of the original issue stands.

## Should `to_jax` take a `device=`?

Worth deciding rather than leaving implicit. Arguments both ways:

- **For**: symmetry with `to_torch`, and it removes a footgun that costs silent throughput.
- **Against**: torch's `device=` earns its place by page-locking the batch buffers and issuing
  the copy against the pool's own lifetime tracking; a JAX `device=` would be a bare
  `device_put` the caller can write themselves, and adding it means a second way to do
  something JAX already does well.

I lean against the parameter and towards the documentation, which is what has shipped — but
the throughput trap is real enough that it should be a deliberate call.

## Repro

```bash
uv venv /tmp/venv-jaxcuda --python 3.12
uv pip install --python /tmp/venv-jaxcuda/bin/python "jax[cuda12]" .
# then the fixture + loop above; script kept at design-notes/gpu-precheck-2026-09-11/
```

Note `uv pip install` does **not** honour `UV_PROJECT_ENVIRONMENT` — pass `--python` explicitly,
or you will install JAX beside torch in the project venv and hit the documented segfault hazard.

emfdavid:
## Summary

Closes #88.

`to_jax` was `jnp.from_dlpack(array)`. That imports a **host** buffer, so the result is
committed to `cpu:0` — and `jax.jit` does not move a committed array. Every JAX user with a
GPU therefore got a pipeline that was correct, silent, and entirely on the wrong device.

Measured on the L4 (`jax[cuda12]`, jax 0.11.1, `jax.devices() == [CudaDevice(id=0)]`):

| | before | after |
|---|---|---|
| `to_jax(batch)["t2m"].devices()` | `[CpuDevice(id=0)]` | `[CudaDevice(id=0)]` |
| `jax.jit(lambda x: x*2)(arr).devices()` | `{CpuDevice(id=0)}` | `{CudaDevice(id=0)}` |
| values vs store | exact | exact |

The batch is now placed on `jax.devices()[0]` — where every other jax array-creation path
puts one, so `from_dlpack` was the anomaly rather than this being a new policy. `device=`
overrides it.

**Why this needs no new machinery.** On a CPU-only build the put is a pass-through that still
aliases the exported buffer (verified: mutating the host array after `device_put` is still
visible through the result), so it costs nothing there. On GPU, `device_put` is
documented-async, but JAX keeps the source *referenced* across the transfer, so the pool's
existing liveness poll defers reclaim — this is the property DESIGN.md already establishes
under "the pool is sound for every adapter". It does **not** make JAX own its transfer the
way `to_torch(device=)` does; JAX cannot consume pinned host memory
([jax#22346](https://github.com/jax-ml/jax/issues/22346)) and exposes no event to gate
recycling on, so **pinning stays torch-only** and that reasoning is unchanged.

Second, smaller fix: bare `jax>=0.4.30` resolves to the **CPU wheel**, so `--extra jax` on a
CUDA box installed a JAX that could not see the GPU at all — before `to_jax` was even
reached. A new **`jax-cuda`** extra pins `jax[cuda12]`, kept separate so `--extra jax` stays
resolvable on any machine (CI depends on that).

## For reviewers

**The thing I would most value a second look at is the regression test's reach.** It asserts
against `jax.devices()[0]` rather than a literal, which makes it meaningful on a CUDA box and
still a contract guard on a CPU one — but it follows that **on CI it would not have caught
this defect**, because CI has no GPU runner and `devices()[0]` is `cpu:0` there. The test
passed before this change on CPU and failed before it on the L4. I verified the failing-then-
passing transition manually on the GPU box in both a `jax[cuda12]` venv and a CPU-only one
(3 passed in each); I could not make CI prove it. If you want a stronger guard, the honest
options are a GPU runner or asserting the narrower property that `to_jax` never returns an
array committed to a device other than the default — neither of which I have added here.

I am confident in the safety argument, because it is not new: it is the referenced-source
property DESIGN.md:624-690 already relies on for the existing zero-copy path, and
`bench/probe_batch_buffers.py --arms all` plus `test_probe_jax_arm_runs_without_a_gpu` are the
standing instruments for it. I did not touch `ChunkPool`, the scheduler, or any cross-thread
readiness path.

One behaviour change worth noting explicitly: `to_jax` is no longer guaranteed to hand back a
host array. Anyone relying on that (`np.asarray` on the result still works, and copies from
device) would see a transfer they did not before. I think that is the right default given the
alternative is silently training on the wrong device, but it is a judgement call rather than a
pure bug fix.

## Author attestation

**Left unchecked deliberately** — this description and the code were drafted by an assistant,
and that box is the human author's to tick.

## Checklist

- [x] Tests added or updated — with the reach caveat described above: the new tests fail
      before this change on a CUDA box and pass on a CPU-only one, so they do not reproduce
      the bug in CI
- [x] `ruff check`, `ruff format --check`, `mypy` and `pytest -q` green locally
      (668 passed, 9 skipped), and `mkdocs build --strict` too
- [x] Docstrings and API docs for any new or changed public surface (`to_jax` flows into
      `docs/api.md` via mkdocstrings)
- [x] User-facing behavior documented in `docs/*.md` (`jax-cuda` added to the extras lists in
      `docs/index.md` and `docs/contributing.md`)
- [x] No load-bearing invariant is broken — `Batch` is still numpy, the hot path is untouched,
      and this is adapter-only

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,31 @@
 
 ## Unreleased
 
+- **A JAX run on a GPU box trained on the CPU, at exact values, with nothing raised.**
+  `to_jax` was `jnp.from_dlpack(array)`, which imports a *host* buffer and therefore commits
+  its result to `cpu:0` — and `jax.jit` does not move a committed array. So every JAX user
+  with a GPU got a pipeline that was correct, silent, and entirely on the wrong device;
+  measured on an L4, `to_jax` returned `cpu:0` with `jax.devices() == [CudaDevice(id=0)]`, and
+  `jit(x * 2)` on that array stayed on the CPU. Nothing in the docs said so, because the
+  adapter had no device story at all: `device=` exists on `as_torch`/`to_torch` only, and
+  `to_jax` took no such argument.
+
+  The batch is now placed on `jax.devices()[0]`, which is where every other jax
+  array-creation path puts one — `jnp.from_dlpack` was the anomaly, not the fix. Pass
+  `device=` to choose another. On a CPU-only build the put is a **pass-through that still
+  aliases** the exported buffer, so it costs nothing there, and the transfer needs no new
+  machinery on GPU: JAX keeps the source referenced across an async `device_put`, so the
+  pool's liveness poll already defers reclaim (DESIGN.md, "the pool is sound for every
+  adapter"). This does not make JAX own its transfer the way `to_torch(device=)` does —
+  JAX cannot consume pinned host memory ([jax#22346](https://github.com/jax-ml/jax/issues/22346))
+  and exposes no event to gate recycling on, so pinning stays torch-only.
+
+- **`--extra jax` on a CUDA box installed a JAX that could not see the GPU.** Bare
+  `jax>=0.4.30` resolves to the CPU wheel, so the one command a GPU user would run left them
+  on the CPU before `to_jax` was even reached. A new **`jax-cuda`** extra pins `jax[cuda12]`.
+  It is a separate extra rather than a change to `jax` so `--extra jax` stays resolvable on
+  any machine, which CI depends on; one framework per environment still applies.
+
 - **The batch-buffer figures in the epoch line were the pool's, printed under one split's name
   (#84).** One buffer pool serves every iteration open on a dataset, and its counters were reset
   at each pass's *start* -- so with `zip(ds.train, ds.val)` a pass that drew 7 batches reported
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -109,7 +109,8 @@ Extras, one at a time (see the framework caveat below):
 
 ```bash
 uv sync --extra torch      # torch handoff (frameworks.as_torch)
-uv sync --extra jax        # JAX handoff (frameworks.to_jax)
+uv sync --extra jax        # JAX handoff (frameworks.to_jax) -- CPU wheel
+uv sync --extra jax-cuda   # the same, on a CUDA box (jax[cuda12])
 uv sync --extra tf         # TF handoff (frameworks.as_tf_dataset)
 uv sync --extra bench      # benchmark suite (xbatcher baseline + plotly + py-spy)
 uv sync --extra docs       # MkDocs site
--- a/docs/index.md
+++ b/docs/index.md
@@ -124,7 +124,8 @@ uv run python -m examples.wb2_dataloader \
 ```bash
 uv sync                  # core engine + dev tools
 uv sync --extra torch    # torch handoff (frameworks.as_torch)
-uv sync --extra jax      # JAX handoff (frameworks.to_jax)
+uv sync --extra jax      # JAX handoff (frameworks.to_jax) -- CPU wheel
+uv sync --extra jax-cuda # the same, on a CUDA box (jax[cuda12])
 uv sync --extra tf       # TF handoff (frameworks.as_tf_dataset)
 uv sync --extra bench    # benchmark suite (xbatcher baseline + plotly)
 uv sync --extra gpu      # CUDA box only: cupy + kvikio zero-copy path
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,6 +60,16 @@ jax = [
     "flax>=0.8",
     "optax>=0.2",
 ]
+# JAX on an NVIDIA GPU. Bare `jax` resolves to the *CPU* wheel, so `--extra jax` on a CUDA
+# box yields a JAX that cannot see the GPU at all -- silently, since nothing fails. This
+# pins the CUDA build instead; it is a separate extra rather than a change to `jax` so that
+# `--extra jax` stays resolvable everywhere, which CI depends on. One framework per
+# environment still applies.
+jax-cuda = [
+    "jax[cuda12]>=0.4.30",
+    "flax>=0.8",
+    "optax>=0.2",
+]
 tf = [
     "tensorflow>=2.16",
 ]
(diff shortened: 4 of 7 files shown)


emfdavid (commented):
:shipit:

emfdavid:
Verified on the GPU box, through the new extra rather than by copying files in — including a
pre-fix control run in the same session.

## The extra resolves

`uv sync --extra jax-cuda` gives **jax / jaxlib 0.10.2** with `jax.devices() ==
[CudaDevice(id=0)]`, and neither torch nor TensorFlow present, so one-framework-per-environment
holds. The project's own `.venv` (torch) was confirmed untouched afterwards.

## Tests

| | result |
|---|---|
| `tests/test_jax.py`, CUDA JAX | **3 passed** — here `test_to_jax_lands_on_the_default_device` is asserting `CudaDevice(id=0)`, which is the case CI cannot reach |
| full suite, `jax-cuda` env | **622 passed, 26 skipped** — every skip an absent optional extra (cloudpickle / torch / icechunk) or `--remote` |

## The defect, and the control

The deterministic evidence first, since it needs no clock. On **pre-fix `main`**, in the same
CUDA-capable environment:

```
jax.devices(): [CudaDevice(id=0)] | to_jax -> [CpuDevice(id=0)]
```

Then `examples/advection/train_jax.py`, with a pre-fix control measured in the same session:

| run | commit | `--device` | persistence RMSE | model RMSE | wall |
|---|---|---|---|---|---|
| **control** | `main`, pre-fix | `cuda` | 0.884 | 0.596 | **27.36 s** |
| this PR | `b1a5cee` | `cpu` | 0.884 | 0.596 | 27.33 s |
| this PR | `b1a5cee` | `cuda` | 0.884 | 0.596 | **9.83 s** |

**Pre-fix, `--device cuda` ran 0.1% away from the CPU arm** (27.36 vs 27.33 s). The flag was
inert — the GPU was never touched, which is the whole defect. After the fix the same arm is
9.83 s.

`persistence RMSE` is **0.884 in all three runs**, as are model RMSE and train MSE (0.3261).
That is the example's own model-independent fingerprint: the fix moved where the compute
happens without changing a byte the loader returns.

Incidentally this makes `train_jax.py`'s existing docstring true — it already claimed "JAX
auto-places on the accelerator it finds (install `jax[cuda12]` for GPU)", which it did not.

## Setup

- **Hardware:** GCP `g2-standard-16`, 16 vCPU, 62 GB RAM, 1x NVIDIA L4 (23 GB), driver
  580.173.02, `us-central1-a`.
- **Store:** the example's default **synthetic** store on local disk (`file://` temp path) —
  `n_steps=768`, `size=48`, `sample_chunk=48`, no inner tiling (one chunk over the whole
  48x48 field), 3 variables, ~21 MB total. Page-cache resident, so this is compute-bound and
  not an IO measurement.
- **Loader config:** defaults — `batch_size=32`, `--epochs 2`, no `cache_dir`.
- **Control:** the pre-fix `main` row above, same box, same env, same session.

## Caveats, stated rather than implied

- **This is not a throughput claim about insitubatch.** The wall times are evidence for one
  binary question — is the GPU being used — and the CPU arm is the control for exactly that.
  One run per arm, no bracketing or interleaving, so treat the times as coarse. The device
  probe above is the primary evidence and is deterministic.
- The control ran on the box's `main` at `1411c54`, which is behind `origin/main` (`d91dcec`).
  I checked: `src/insitubatch/frameworks.py` and `examples/advection/train_jax.py` are
  **byte-identical** across those two commits, so the control is valid for this question. The
  14 files that differ are elsewhere.
- Still true, and unchanged by any of the above: **CI cannot catch a regression here**, because
  `jax.devices()[0]` is `cpu:0` without a GPU runner. That limitation is in the PR description.

PR #90, opened 2026-09-11
docs: rewrite the README around operating the library, organize the changelog, and call the project Beta

emfdavid:
Three separable commits; they read best in order.

| commit | what |
|---|---|
| `91d4eb7` | README rewrite, and Beta across all three places that state maturity |
| `42bb60d` | changelog grouped by kind, with an `### Upgrading` preamble |
| `668df8f` | a written release process, which did not exist |

## Summary

**The README was a strong pitch and a weak manual, and we measured it.** Two cold agents, no
prior context, each given an unfamiliar public store and one documentation surface — README
alone, or README plus `docs/tuning.md` — and asked to build a working iterator. Both got
there. Neither got there from the page: the README-only arm made **sixteen** signature and
docstring inspections, every one driven by a question the page did not answer. Both arms hit
the same wall in the same place on *different* docs and *different* datasets, which is what
makes it the page's defect rather than one agent's bad luck.

The API section began 54% of the way down. It now begins at 18%, and what is there is
correct. Previously: nothing said how to select variables, so both agents opened every array
in their store and were quoted residency budgets of **5,493 GiB** and **10,935 GiB** for
variables they never asked for; the transform parameters were named in the singular four times
in bold and never shown being passed, so the only place the names appeared was the place that
got them wrong; `applies(...)` was absent entirely, and with it the silent-truncation warning
that lives only in its docstring; `sample_range`, `readonly_cache` and `reset_stale_cache` did
not appear at all, and `cache_dir` appeared once, in the Windows row of the platform table.

Caching and diagnostics now have sections of their own. The first is this release's headline
behaviour and had a single mention; the second (`ds.last_pass`, `limiting_stage`,
`print_debug_info()`) had none. The free-threading essay and the milestone inventory are gone
— `DESIGN.md` is the single source of truth for status, and the README was mirroring it.

**Both executable blocks are now verified by execution, which is the only thing that caught
what was wrong with them.** The quickstart could not run at all (`n_epochs` was undefined) and
asked for ~16 GB of reads; it is a runnable multi-variable recipe over a ~2 GB window,
extracted from the file and executed. The `print_summary()` transcript had numbers carried
over from a different configuration; it is regenerated from a real run of the dataset the page
builds. The `check-transform` invocation used an undefined `$STORE_URL`; it now names the
public WeatherBench2 store and was run as written. And a claim that a non-zero exit "gates a
pre-commit hook" was simply false — no such hook is shipped — which is now stated as what you
*can* do rather than what we do.

**Status becomes Beta.** The three statements disagreed: the PyPI classifier said
`2 - Pre-Alpha`, the README said alpha, `DESIGN.md` said Alpha. Beta is claimed on
instrumentation, not on an absence of bugs — the O(chunks) invariant is pinned by a test with
a companion control that fails loudly if the counter comes unwired, byte fingerprints and a
model-free persistence baseline cover the silently-plausible-data class, and the CUDA-gated
buffer-lifetime tests now run green on an L4 with the instrument validated first. Still
pre-1.0; breaking changes still allowed.

**The changelog** had grown to 791 lines across 51 flat entries — 86% of the file, no
subheadings anywhere. Grouped under six headings, **move-only**: every entry byte-identical,
the multiset of content lines unchanged. Contributors keep appending flat bullets under
`## Unreleased` with no category to guess at; sorting is a release-cut step. The one addition
is `### Upgrading`, because the file's real defect was a landmine, not navigation: the
manifest bump that invalidates every existing `cache_dir` sat as a bolded clause five lines
into a fifteen-line entry about tile-major storage, with a second bump to format 4 in a
different entry again.

**The release process** did not exist in writing anywhere. That is why the docs-usability test
had no home despite being intended to run before every release, and how three maturity
statements drifted two steps apart unnoticed.

## For reviewers

- **`42bb60d` is move-only and the diff will not show that usefully.** I verified it by
  comparing the multiset of content lines before and after: 0 lost, 9 added, and those 9 are
  the Upgrading preamble. Worth reviewing the *heading assignments*, not the diff — a few
  entries could defensibly sit in a different group.
- **Beta is the judgement call here**, and it is yours. My argument is that the instruments
  now exist and are load-bearing, not that the bugs are gone — 0.2.0 closed nine
  silently-plausible-data defects in one cycle, which is evidence of a high discovery rate as
  much as of diligence.
- **#86 will conflict on rebase, and the resolution is to drop its README hunk.** Its only
  README change adds a `persist=True` clause to the Status paragraph, which this deletes
  outright; that claim now lives in the Caching section, stated more fully.
- I removed every number from the README and let the Benchmarks link carry them, so there are
  no performance claims in this PR to check setup for.

## Author attestation

**Left unchecked deliberately** — the code and this description were drafted by an assistant,
and that box is the human author's to tick.

## Checklist

- [x] `ruff check`, `ruff format --check`, `mypy` and `pytest -q` green locally (668 passed,
      9 skipped), and `mkdocs build --strict` too
- [x] Both runnable README blocks extracted and executed (exit 0); `print_summary()` transcript
      regenerated from a real run; all 19 README links return 200
- [x] User-facing behavior documented in `docs/*.md` — the release process in
      `docs/contributing.md`
- [x] Bullets added under `## Unreleased` in `CHANGELOG.md`
- [x] No load-bearing invariant is broken — no source change in this PR beyond two test
      docstrings that asserted things about the README this makes false
- [x] No performance claim is made

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,47 @@
 
 ## Unreleased
 
+- **The README was a strong pitch and a weak manual, and a cold-agent test measured how weak.**
+  Two agents with no prior context were each given an unfamiliar public store and one
+  documentation surface — README alone, or README plus `docs/tuning.md` — and told to build a
+  working iterator. Both got there; neither got there from the page. The README-only arm made
+  **sixteen** signature and docstring inspections, every one driven by a question the page did
+  not answer, and concluded that "nearly everything I needed to *operate* the library came from
+  parameter signatures, docstrings and error strings". The other arm, on different docs and a
+  different dataset, hit the identical wall in the identical place — which is what makes it the
+  page's defect rather than one agent's bad luck.
+
+  What the page got wrong was concentrated and consequential. The only end-to-end example was
+  single-variable, because the demo store has one array, and nothing said how to select
+  variables — so both agents opened **every** array in their store and were quoted residency
+  budgets of 5 493 GiB and 10 935 GiB for variables they never asked for. The transforms
+  section named `chunk_transform` / `batch_transform` four times in bold; the parameters are
+  plural and take a sequence, and the page never showed either being passed, so the one place
+  the names appeared was the place that got them wrong. `applies(...)` was absent entirely,
+  and with it the warning that lives only in its docstring: gating inside a transform body
+  instead makes unaffected arrays "gathered as truncated prefixes of themselves ... with no
+  exception raised". `sample_range`, `readonly_cache` and `reset_stale_cache` did not appear;
+  `cache_dir` appeared once, in the Windows row of the platform table. And following the page
+  verbatim the loader ran in silence, because nothing mentioned `logging.basicConfig`.
+
+  The page is rewritten around what it is for. The API now starts at 18% of the page rather
+  than 54%; the quickstart is a runnable multi-variable recipe that was executed verbatim to
+  confirm it (it previously could not run at all — `n_epochs` was undefined); caching and
+  diagnostics get sections of their own, the first being this release's headline behaviour and
+  the second — `ds.last_pass`, `limiting_stage`, `print_debug_info()` — having had no README
+  coverage whatsoever. The free-threading essay and the milestone inventory are gone, the
+  latter because `DESIGN.md` is the single source of truth for status and the README was
+  mirroring it.
+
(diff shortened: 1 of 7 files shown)


emfdavid (commented):
:shipit:

